PR #1, opened 2018-11-27
Import old work minus compile-time disabled hardware tracer code.

vext01:
This is going to be horrible to review as it's all one commit. Most of the incoming code here was reviewed in the old repo (now ykrustc.old). Sorry about this, but it's too hard to pick it all apart retrospectively.

As discussed the other day, we decided to start a fresh repo and copy over only what we want for now. All code that had been guarded with `yk_hwt` has been removed, and I've also fixed the tests along the way.

The tests all pass :)

Hopefully bors won't be upset that the repo has been pulled from under it's feet :)

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -1,0 +1,10 @@
+#!/bin/sh
+#
+# Build script for continuous integration.
+
+./x.py clean  # We don't clone afresh to save time and bandwidth.
+git clean -dffx # If upstream removes a submodule, remove the files from disk.
+
+# Note that the gdb must be Python enabled.
+PATH=/opt/gdb-8.2/bin:${PATH} RUST_BACKTRACE=1 YK_DEBUG_SECTIONS=1 \
+    ./x.py test --config .buildbot.toml
--- a/README_YORICK.md
+++ b/README_YORICK.md
@@ -1,0 +1,4 @@
+# Yorick-specific Notes
+
+To enable the experimental `.yk_mir_cfg` ELF section, set `YK_DEBUG_SECTIONS`
+in your environment.
--- a/src/librustc_driver/driver.rs
+++ b/src/librustc_driver/driver.rs
@@ -11,6 +11,7 @@
 use rustc::dep_graph::DepGraph;
 use rustc::hir::{self, map as hir_map};
 use rustc::hir::lowering::lower_crate;
+use rustc::hir::def_id::LOCAL_CRATE;
 use rustc_data_structures::fingerprint::Fingerprint;
 use rustc_data_structures::stable_hasher::StableHasher;
 use rustc_mir as mir;
@@ -33,11 +34,15 @@ use rustc_metadata::creader::CrateLoader;
 use rustc_metadata::cstore::{self, CStore};
 use rustc_traits;
 use rustc_codegen_utils::codegen_backend::CodegenBackend;
+use rustc_codegen_utils::link::out_filename;
 use rustc_typeck as typeck;
 use rustc_privacy;
 use rustc_plugin::registry::Registry;
 use rustc_plugin as plugin;
 use rustc_passes::{self, ast_validation, hir_stats, loops, rvalue_promotion};
+use rustc_yk_sections::mir_cfg::emit_mir_cfg_section;
+use rustc_yk_sections::with_yk_debug_sections;
+use rustc::util::nodemap::DefIdSet;
 use super::Compilation;
 
 use serialize::json;
@@ -50,7 +55,7 @@ use std::io::{self, Write};
 use std::iter;
 use std::path::{Path, PathBuf};
 use rustc_data_structures::sync::{self, Lrc, Lock};
-use std::sync::mpsc;
+use std::sync::{mpsc, Arc};
 use syntax::{self, ast, attr, diagnostics, visit};
 use syntax::early_buffered_lints::BufferedEarlyLint;
 use syntax::ext::base::ExtCtxt;
@@ -325,7 +330,7 @@ pub fn compile_input(
                     tcx.print_debug_stats();
                 }
 
-                let ongoing_codegen = phase_4_codegen(&*codegen_backend, tcx, rx);
+                let (ongoing_codegen, def_ids) = phase_4_codegen(&*codegen_backend, tcx, rx);
 
                 if log_enabled!(::log::Level::Info) {
                     println!("Post-codegen");
@@ -337,6 +342,17 @@ pub fn compile_input(
                         sess.err(&format!("could not emit MIR: {}", e));
                         sess.abort_if_errors();
                     }
+                }
+
+                // Output Yorick debug sections into binary targets.
+                if sess.crate_types.borrow().contains(&config::CrateType::Executable) &&
+                    with_yk_debug_sections() {
+                    let out_fname = out_filename(
+                        tcx.sess, config::CrateType::Executable, &outputs,
+                        &*tcx.crate_name(LOCAL_CRATE).as_str());
+
+                    tcx.sess.yk_link_objects.borrow_mut()
+                       .push(emit_mir_cfg_section(&tcx, &def_ids, out_fname));
                 }
 
                 Ok((outputs.clone(), ongoing_codegen, tcx.dep_graph.clone()))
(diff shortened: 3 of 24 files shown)


vext01:
bors ping

vext01:
bors ping

ltratt:
> Most of the incoming code 

Most? All? How do I know what's what?

vext01:
Shall I point out the test fixes?

ltratt:
Yes please!

.buildbot.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+#
+# Build script for continuous integration.
+
+./x.py clean  # We don't clone afresh to save time and bandwidth.
+
+# Note that the gdb must be Python enabled.
+PATH=/opt/gdb-8.2/bin:${PATH} RUST_BACKTRACE=1 YK_DEBUG_SECTIONS=1 \
```
vext01, line 9: I added the `YK_DEBUG_SECTIONS` env here so that we test the compiler with the `.yk_mir_cfg` section enabled. Also switched to full stage 2 tests.

README_YORICK.md
```diff
@@ -0,0 +1,4 @@
+# Yorick-specific Notes
```
vext01, line 1: This file had spoken about the (now removed) `yk_hwt` compile-time gate. Now it talks about the environment var to enable the MIR ELF section.

src/librustc_driver/driver.rs
```diff
@@ -339,6 +344,17 @@ pub fn compile_input(
                     }
                 }
 
+                // Output Yorick debug sections into binary targets.
+                if sess.crate_types.borrow().contains(&config::CrateType::Executable) &&
+                    with_yk_debug_sections() {
```
vext01, line 349: This block of code was being run even if we didn't request debug sections. I've added the `&& with_yk_debug_sections()` to the condition.

src/librustc_yk_sections/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+#![feature(box_patterns)]
+
+extern crate rustc;
+extern crate rustc_yk_link;
+extern crate rustc_codegen_utils;
+extern crate byteorder;
+
+use std::env;
+
+/// Are Yorick debug sections enabled?
```
vext01, line 19: I added this convenience function for checking if the debug sections are enabled in the env.

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -0,0 +1,214 @@
...
+const MIR_CFG_SECTION_NAME: &'static str = ".yk_mir_cfg";
+const SECTION_VERSION: u16 = 0;
+
+/// Serialises the control flow for the given `DefId`s into a ELF object file and returns a handle
+/// for linking.
+pub fn emit_mir_cfg_section<'a, 'tcx, 'gcx>(
+    tcx: &'a TyCtxt<'a, 'tcx, 'gcx>, def_ids: &DefIdSet, exe_filename: PathBuf)
+    -> YkExtraLinkObject {
+
+    // Serialise the MIR into a file whose name is derived from the output binary. The filename
```
vext01, line 60: This fixes the reproducible build tests. We had used a tempfile with a random name (via `mkstemp`) before.

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -0,0 +1,214 @@
...
+    // must be the same between builds of the same binary for the reproducible build tests to pass.
+    let mut mir_path: String = exe_filename.to_str().unwrap().to_owned();
+    mir_path.push_str(".ykcfg");
+    let mut fh = File::create(&mir_path).unwrap();
+
+    // Write a version field for sanity checking when deserialising.
+    fh.write_u16::<NativeEndian>(SECTION_VERSION).unwrap();
+
+    // To satisfy the reproducible build tests, the CFG must be written out in a deterministic
```
vext01, line 69: This is also needed to fix the reproducible build tests.

src/test/run-make-fulldeps/hotplug_codegen_backend/the_backend.rs
```diff
@@ -44,10 +45,10 @@ impl CodegenBackend for TheBackend {
         &self,
         tcx: TyCtxt<'a, 'tcx, 'tcx>,
         _rx: mpsc::Receiver<Box<Any + Send>>
-    ) -> Box<Any> {
```
vext01, line 47: This test file was checking that you could implement a codegen backend via a trait. I had changed the trait signature in order to collect the `DefId`s that need to appear in the MIR ELF section, so we need to change the test.

vext01 (commented):
I've commented on the aspects which need your attention.

Cheers.

ltratt:
Dumb question: are we still keeping the debug sections in? I thought we were ditching that aspect.

vext01:
The debug sections is currently just the `.yk_mir_cfg` section which contains a serialised version of the MIR CFG. I plan to use this later for tests, checking that the traces we get are a valid path through the MIR CFG. We'd need this to test both hardware and software tracing.

ltratt:
So this isn't the LLVM debug stuff in any way? [Just checking!]

vext01:
That's right. All `DILabel` stuff has been removed completely. I've got all the old DILabel code stashed away in a private repo for future reference if/when the time comes.

ltratt:
OK, so I am fine with everything here.

notriddle:
Bors seems fine, but I think Travis is mad at you.

vext01:
It's the buildbot master. It's importing the whole rustc history from the staging branch at about two commits a second...

vext01:
> tidy error: /home/buildbot-worker2/bencher12-worker2/buildbot-build-script/build/src/jemalloc/test/stress/microbench.c:132: tab character

I've seen this error before. It's because jemalloc was removed from git, but the pull doesn't remove the code (and it is this code which fails the tidy linter).

We could try adding a `git clean -df` step into the CI script, or we could clone afresh for each build.

(About the buildbot master database, it's 0.5GiB in size -- I'll ask around to see if that's expected)

Cheers

ltratt:
> It's because jemalloc was removed from git, but the pull doesn't remove the code (and it is this code which fails the tidy linter).

Do you mean, doesn't remove the object files and so on?

vext01:
> Do you mean, doesn't remove the object files and so on?

If I see it correctly, the pull doesn't remove the source code on-disk (notice the error is talking about a tab in`microbench.c`). The files end up in unstaged state, but still get linted.

ltratt:
This sounds like a one-off problem? If so, we could just delete the whole cache (I assume) and restart?

vext01:
> This sounds like a one-off problem?

I'd like it to be fully automated if possible.

FWIW, here's the worker's git clone status:
```
buildbot-worker2@bencher12:~/bencher12-worker2/buildbot-build-script/build$ git status
On branch staging
Untracked files:
  (use "git add <file>..." to include in what will be committed)

        src/jemalloc/
```

Adding a `git rm -df` to the build script would have killed that.

Also we could just clone afresh every time. Since we are doing an `x.py clean` all we save is bandwidth at this point.

ltratt:
>>  This sounds like a one-off problem?
> I'd like it to be fully automated if possible.

If it's a one-off problem, automating it might be a *bad* idea, because future instances of this might not have the same solution. The problem with `git clean` is that it's likely to remove files that we want to stick around to improve build times.

vext01:
> If it's a one-off problem

It's a problem every time upstream removes a submodule.

> The problem with git clean is that it's likely to remove files that we want to stick around to improve build times.

`x.py clean` is already blasting any useful files.

ltratt:
> x.py clean is already blasting any useful files.

Ah, OK, then let's do `git clean`. I'd prefer to avoid downloading everything every time if possible: it's slow and puts a strain on other people.

vext01:
>  I'd prefer to avoid downloading everything every time if possible

Agreed! Let's give it a shot.

vext01:
There's the commit. Do we want to squash it in?

ltratt:
No, we need to try it first.

bors try

vext01:
> No, we need to try it first.

How comes? Just so we can revert it easily if it doesn't work?

ltratt:
Yes, for example. But, in general, I don't think we should do an rplus until we have a fairly high degree of confidence that it'll pass. If we don't, we should do a try.

ltratt:
Well, that was quick!

vext01:
> tidy check
tidy error: /home/buildbot-worker1/bencher12-worker1/buildbot-build-script/build/src/jemalloc/msvc/projects/vc2015/test_threads/test_threads.cpp:29: line longer than 100 chars

Well I'm confused. Looking into it.

vext01:
> + git clean -df
> Skipping repository src/jemalloc/

Doh!

vext01:
I found [here](https://github.com/gitlabhq/gitlab-ci-runner/pull/141) someone using `git clean -ffdx` for CI.

>        -f, --force
>           If the Git configuration variable clean.requireForce is not set to
>           false, git clean will refuse to delete files or directories unless
>           given -f, -n or -i. Git will refuse to delete directories with .git
>           sub directory or file unless a second -f is given.

>        -x
>           Don’t use the standard ignore rules read from .gitignore (per
>           directory) and $GIT_DIR/info/exclude, but do still use the ignore
>           rules given with -e options. This allows removing all untracked
>           files, including build products. This can be used (possibly in
>           conjunction with git reset) to create a pristine working directory
>           to test a clean build.

So we can try adding a second `-f` and also a `-x`. OK with that?

vext01:
(The reason it doesn't kill `src/jemalloc` is because even though it's untracked, it contains a `.git`)

ltratt:
Let's try it.

vext01:
Like so. Give bors a kick.

ltratt:
I think you should also feel free to do "try" with bors. Ultimately you can do whatever you want except a) squash b) merge.

vext01:
bors try

vext01:
That did the trick.

```
...
+ git clean -dffx
Removing build/
Removing src/bootstrap/__pycache__/
Removing src/bootstrap/bootstrap.pyc
Removing src/etc/__pycache__/
Removing src/jemalloc/
...
tidy check
* 568 error codes
* highest error code: E0721
* 239 features
Building stage0 std artifacts (x86_64-unknown-linux-gnu -> x86_64-unknown-linux-gnu)
   Compiling cc v1.0.25
   Compiling core v0.0.0 (/home/buildbot-worker1/bencher12-worker1/buildbot-build-script/build/src/libcore)
...
```

ltratt:
Looks like it's time to squash!

vext01:
Splat!

PR #2, opened 2018-12-31
The start of software tracing.

vext01:
Here's the long-awaited PR for the beginnings of software tracing.

See the commit messages for a description of the changes.

Notes:
 - The body of the software tracing function is not yet implemented. This will come next. I plan for it to store the locations it receives into a thread local vector. A function will then be implemented to acquire the contents.

 - I've decided not to test any of our new behaviours in existing tests, as we should eventually have our own tests if we decide to keep this code. I will raise an issue if/when we merge.

 - At some point in the future, we will have to have a compile-time switch to choose between software or hardware tracing. We would need some kind of conditional ignore flag for tests. Any test I've marked `FIXME ignore_swt` can be run if software tracing was not built in. Note that the choice between hardware or software tracing will need to be compile-time, as the standard library must have been annotated if we want software tracing. In a hardware tracing scenario, we don't want the overhead of the calls to the software trace function.

Comments? Questions? I expect there will be many.

(Happy new year, for later!)

### DIFF
--- a/src/librustc_mir/transform/add_yk_swt_calls.rs
+++ b/src/librustc_mir/transform/add_yk_swt_calls.rs
@@ -1,0 +1,186 @@
+// Copyright 2018 King's College London.
+// Created by the Software Development Team <http://soft-dev.org/>.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use rustc::ty::{self, TyCtxt, List};
+use rustc::mir::{Operand, LocalDecl, Place, SourceInfo, BasicBlock, Local, BasicBlockData,
+    TerminatorKind, Terminator, OUTERMOST_SOURCE_SCOPE, Constant, Mir};
+use rustc_data_structures::indexed_vec::Idx;
+use syntax_pos::DUMMY_SP;
+use syntax::attr;
+use transform::{MirPass, MirSource};
+use rustc::hir;
+use rustc::hir::def_id::{DefIndex, LOCAL_CRATE};
+use rustc::hir::map::blocks::FnLikeNode;
+
+/// A MIR transformation that, for each basic block, inserts a call to the software trace recorder.
+/// The arguments to the calls (crate hash, DefId and block index) identify the position to be
+/// inserted into a trace.
+///
+/// The transformation works by copying each original "user block" and replacing it with a new
+/// block -- its "shadow" -- which calls the trace recorder function, before returning to the copy.
+///
+/// For example:
+///
+/// +----+   +----+           +-----+   +-----+   +----+   +-----+   +-----+   +----+
+/// | B0 |-->| B1 |  Becomes: | B0' |-->| Rec |-->| B2 |-->| B1' |-->| Rec |-->| B3 |
+/// +----+   +----+           +-----+   +-----+   +----+   +-----+   +-----+   +----+
+///
+/// Where:
+///  * B0 and B1 are "user blocks" in the MIR before the transformation.
+///  * B0' and B1' are "shadow blocks" of B0 and B1 respectively.
+///  * B2 and B3 are copies of B0 and B1 respectively.
+///  * 'Rec' is the trace recorder function.
+///  * The block indices match the indices in the backing vector in the MIR.
+///
+/// The extra calls we insert mean that we have to allocate new local decls for the (unit) return
+/// values: one new decl for each call.
+pub struct AddYkSWTCalls(pub DefIndex);
+
+impl MirPass for AddYkSWTCalls {
+    fn run_pass<'a, 'tcx>(&self,
+                          tcx: TyCtxt<'a, 'tcx, 'tcx>,
+                          src: MirSource,
+                          mir: &mut Mir<'tcx>) {
+        if is_untraceable(tcx, src) {
+            return;
+        }
+
+        let rec_fn_defid = tcx.get_lang_items(LOCAL_CRATE).yk_swt_rec_loc_wrap()
+            .expect("couldn't find software trace recorder function");
+
+        let unit_ty = tcx.mk_unit();
+        let u32_ty = tcx.types.u32;
+        let u64_ty = tcx.types.u64;
+
+        let mut shadow_blks = Vec::new();
+        let mut user_blks = Vec::new(); // Copies of the blocks we started with.
+        let mut new_local_decls = Vec::new();
+
+        let num_orig_local_decls = mir.local_decls.len();
+        let local_crate_hash = tcx.crate_hash(LOCAL_CRATE).as_u64();
+
+        for (bb, bb_data) in mir.basic_blocks().iter_enumerated() {
+            // Copy the block.
+            let new_blk = bb_data.clone();
+            let new_blk_idx = BasicBlock::new(mir.basic_blocks().len() + user_blks.len());
+            user_blks.push(new_blk);
+
+            // Prepare to call the recorder function.
+            let ret_val = LocalDecl::new_temp(unit_ty, DUMMY_SP);
+            let ret_place = Place::Local(Local::new(num_orig_local_decls + new_local_decls.len()));
+            new_local_decls.push(ret_val);
+
+            let crate_hash_oper = Operand::Constant(box Constant {
+                span: DUMMY_SP,
+                ty: u64_ty,
+                user_ty: None,
+                literal: ty::Const::from_u64(tcx, local_crate_hash),
+            });
+
+            let def_idx_oper = Operand::Constant(box Constant {
+                span: DUMMY_SP,
+                ty: u32_ty,
+                user_ty: None,
+                literal: ty::Const::from_u32(tcx, self.0.as_raw_u32()),
+            });
+
+            let bb_oper = Operand::Constant(box Constant {
+                span: DUMMY_SP,
+                ty: u32_ty,
+                user_ty: None,
+                literal: ty::Const::from_u32(tcx, bb.index() as u32),
+            });
+
+            let rec_fn_oper = Operand::function_handle(tcx, rec_fn_defid,
+                List::empty(), DUMMY_SP);
+
+            let term_kind = TerminatorKind::Call {
+                func: rec_fn_oper,
+                args: vec![crate_hash_oper, def_idx_oper, bb_oper],
+                destination: Some((ret_place, new_blk_idx)), // Return to the copied block.
+                cleanup: None,
+                from_hir_call: false,
+            };
+
+            // Build the replacement block with the new call terminator.
+            let source_info = bb_data.terminator.clone().map(|t| t.source_info)
+                .or(Some(SourceInfo { span: DUMMY_SP, scope: OUTERMOST_SOURCE_SCOPE })).unwrap();
+            let replace_block = BasicBlockData {
+                statements: vec![],
+                terminator: Some(Terminator {
+                    source_info,
+                    kind: term_kind
+                }),
+                is_cleanup: false
+            };
+            shadow_blks.push(replace_block);
+        }
+
+        // Finally, commit our transformations.
+        mir.basic_blocks_mut().extend(user_blks);
+        mir.local_decls.extend(new_local_decls);
+        for (bb, bb_data) in shadow_blks.drain(..).enumerate() {
+            mir.basic_blocks_mut()[BasicBlock::new(bb)] = bb_data;
+        }
+    }
+}
+
+/// Given a `MirSource`, decides if it is possible for us to trace (and thus whether we should
+/// transform) the MIR. Returns `true` if we cannot trace, otherwise `false`.
+fn is_untraceable(tcx: TyCtxt<'a, 'tcx, 'tcx>, src: MirSource) -> bool {
+    // Never annotate anything annotated with the `#[no_trace]` attribute. This is used on tests
+    // where our pass would interfere and on the trace recorder to prevent infinite
+    // recursion.
+    //
+    // "naked functions" can't be traced because their implementations manually implement
+    // binary-level function epilogues and prologues, often using in-line assembler. We can't
+    // automatically insert our calls into such code without breaking stuff.
+    for attr in tcx.get_attrs(src.def_id).iter() {
+        if attr.check_name("no_trace") {
+            return true;
+        }
+        if attr.check_name("naked") {
+            return true;
+       }
+    }
+
+    // Similar to `#[no_trace]`, don't transform anything inside a crate marked `#![no_trace]`.
+    for attr in tcx.hir.krate_attrs() {
+        if attr.check_name("no_trace") {
+            return true;
+        }
+    }
+
+    // We can't call the software tracing function if the crate doesn't depend upon libcore because
+    // that's where the entry point to the trace recorder function lives.
+    if attr::contains_name(tcx.hir.krate_attrs(), "no_core") {
+        return true;
+    }
+
+    // Attempting to transform libcompiler_builtins leads to an undefined reference to the trace
+    // recorder wrapper `core::yk_swt::yk_swt_rec_loc_wrap`. It's not worth investigating, as this
+    // crate only contains wrapped C and ASM code that we can't transform anyway.
+    if tcx.is_compiler_builtins(LOCAL_CRATE) {
+        return true;
+    }
+
+    // We can't transform promoted items, because they are `const`, and our trace recorder isn't.
+    if let Some(_) = src.promoted {
+        return true;
+    }
+
+    // For the same reason as above, regular const functions can't be transformed.
+    let node_id = tcx.hir.as_local_node_id(src.def_id)
+        .expect("Failed to get node id");
+    if let Some(fn_like) = FnLikeNode::from_node(tcx.hir.get(node_id)) {
+        fn_like.constness() == hir::Constness::Const
+    } else {
+        true
+    }
+}
--- a/.buildbot.toml
+++ b/.buildbot.toml
@@ -1,5 +1,8 @@
 # Config file for continuous integration.
 
 [rust]
-# Use threads.
-codegen-units = 0
+codegen-units = 0           # Use many compilation units.
+debug-assertions = true     # Turn on assertions in rustc.
+
+[llvm]
+assertions = true           # Turn on assertions in LLVM.
--- a/src/libcore/lib.rs
+++ b/src/libcore/lib.rs
@@ -229,6 +229,8 @@ mod nonzero;
 mod tuple;
 mod unit;
 
+mod yk_swt;
+
 // Pull in the `coresimd` crate directly into libcore. This is where all the
 // architecture-specific (and vendor-specific) intrinsics are defined. AKA
 // things like SIMD and such. Note that the actual source for all this lies in a
--- a/src/libcore/yk_swt.rs
+++ b/src/libcore/yk_swt.rs
@@ -1,0 +1,26 @@
+// Copyright 2018 King's College London.
+// Created by the Software Development Team <http://soft-dev.org/>.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+/// The software trace recorder function.
+/// This is a weak language item, it actually resides in libstd. It has to be weak to allow libcore
+/// to call up to libstd (libstd is not a dependency of libcore).
+extern "Rust" {
+    #[cfg_attr(not(stage0), lang="yk_swt_rec_loc")]
+    fn yk_swt_rec_loc(crate_hash: u64, def_idx: u32, bb: u32);
+}
+
+/// Wrapper lang item to call the above wrapper function.
+/// This has to be a lang item too, as a MIR terminator cannot call a weak language item directly.
+#[allow(dead_code)] // Used only indirectly in a MIR pass.
+#[cfg_attr(not(stage0), lang="yk_swt_rec_loc_wrap")]
+#[cfg_attr(not(stage0), no_trace)]
+fn yk_swt_rec_loc_wrap(crate_hash: u64, def_idx: u32, bb: u32) {
+    unsafe { yk_swt_rec_loc(crate_hash, def_idx, bb) };
+}
+
--- a/src/librustc/middle/lang_items.rs
+++ b/src/librustc/middle/lang_items.rs
@@ -376,6 +376,9 @@ language_item_table! {
     I128ShroFnLangItem,          "i128_shro",          i128_shro_fn,            Target::Fn;
     U128ShroFnLangItem,          "u128_shro",          u128_shro_fn,            Target::Fn;
 
+    YkSwtRecLocLangItem,         "yk_swt_rec_loc",     yk_swt_rec_loc,          Target::Fn;
+    YkSwtRecLocWrapLangItem,     "yk_swt_rec_loc_wrap",yk_swt_rec_loc_wrap,     Target::Fn;
+
     // Align offset for stride != 1, must not panic.
     AlignOffsetLangItem,         "align_offset",       align_offset_fn,         Target::Fn;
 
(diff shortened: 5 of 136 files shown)


vext01:
bors try

vext01:
Wait, the order of the commits is wrong.

vext01:
False alarm. Github's web interface orders the commits by date, not by the order in git.

The order is correct in git, i.e. the test change comes after the MIR pass change. I only committed the tests first to make the output of `git status` readable. There are so many test changes!

bors try

ltratt:
Can you explain `no_trace` vs `yk_swt_rec_loc`? Also, would I be an idiot if I couldn't see either being defined in https://github.com/softdevteam/ykrustc/pull/2/commits/74a44dc00adb50ad16b5d6df944e721531fc80ce? I think that commit is in the wrong place in the list?

src/librustc_mir/transform/add_yk_swt_calls.rs
```diff
@@ -0,0 +1,170 @@
...
+use rustc_data_structures::indexed_vec::Idx;
+use syntax_pos::DUMMY_SP;
+use syntax::attr;
+use transform::{MirPass, MirSource};
+use rustc::hir;
+use rustc::hir::def_id::{DefIndex, LOCAL_CRATE};
+use rustc::hir::map::blocks::FnLikeNode;
+
+/// A MIR pass which, for each basic block, inserts a call to the software trace recorder.
+/// The call arguments passed uniquely identify the MIR location.
```
ltratt: What does the second sentence mean?

src/librustc_mir/transform/add_yk_swt_calls.rs
```diff
@@ -0,0 +1,170 @@
...
+
+        // Types.
+        let unit_ty = tcx.mk_unit();
+        let u32_ty = tcx.types.u32;
+        let u64_ty = tcx.types.u64;
+
+        // Each block is replaced by a new block whose terminator calls the recorder function.
+        let mut replace_blocks = Vec::new();
+
+        // The original blocks are copied and the recorder function returns to a copy.
```
ltratt: Not sure I understand this comment.

src/librustc_mir/transform/add_yk_swt_calls.rs
```diff
@@ -0,0 +1,170 @@
...
+        let mut copied_blocks = Vec::new();
+
+        // New local decls are required to accomodate the (unit) return value of the recorder func.
```
ltratt: Why?

src/librustc_mir/transform/add_yk_swt_calls.rs
```diff
@@ -0,0 +1,170 @@
...
+        let mut new_local_decls = Vec::new();
+
+        let num_orig_blocks = mir.basic_blocks().len();
```
ltratt: Is it useful to pull these out as local vars? I think it probably just makes it harder to read the code below in this case.

src/librustc_mir/transform/add_yk_swt_calls.rs
```diff
@@ -0,0 +1,170 @@
...
+        mir.local_decls.extend(new_local_decls);
+        for (bb, bb_data) in replace_blocks.drain(..).enumerate() {
+            mir.basic_blocks_mut()[BasicBlock::new(bb)] = bb_data;
+        }
+    }
+}
+
+/// Given a `MirSource`, decides if we should annotate the correpsonding MIR.
+fn should_annotate(tcx: TyCtxt<'a, 'tcx, 'tcx>, src: MirSource) -> bool {
+    // Never annotate any MIR-like thing marked `#[no_trace]` or `#[naked]`. The trace record and
```
ltratt: Is it worth briefly explaining why `naked` things aren't traced?

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: The function accepts three arguments which identify the basic block: crate hash, def_id, block_index. Shall I write that there instead?

src/librustc_mir/transform/add_yk_swt_calls.rs
ltratt: I'm not sure I understand the second sentence: I can't see anything in here that marks anything as `no_trace`?

src/librustc_mir/transform/add_yk_swt_calls.rs
```diff
@@ -0,0 +1,170 @@
...
+        // Finally, commit our transformations.
+        mir.basic_blocks_mut().extend(copied_blocks);
+        mir.local_decls.extend(new_local_decls);
+        for (bb, bb_data) in replace_blocks.drain(..).enumerate() {
+            mir.basic_blocks_mut()[BasicBlock::new(bb)] = bb_data;
+        }
+    }
+}
+
+/// Given a `MirSource`, decides if we should annotate the correpsonding MIR.
```
ltratt: Should this be "Given a `MirSource`, return `true` if the corresponding MIR should be annotated or false otherwise." Although I don't think "annotate" is the right word here: we really mean something like "should_be_traced"?

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: We are replacing each block with one that makes a call to the tracing function before returning to a copy of the original block. We have to do it this way to preserve the edge structure of the CFG.

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: We are forced to allocate space for the return value of the call, even though it's `()`. That's just the way it is AFAICS.

src/librustc_mir/transform/add_yk_swt_calls.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    // If there is a crate level `#![no_trace]` attribute, honour that.
+    for attr in tcx.hir.krate_attrs() {
+        if attr.check_name("no_trace") {
+            return false;
+        }
+    }
+
+    // We can't call the software tracing function if there is no libcore.
```
ltratt: This comment, and the rest of the comments in this function, state what the code does. The code is simple enough that the comments aren't very useful in that capacity. I'm wondering if we need to state *why* the code is here for these things? e.g. we might say "We can't trace a function if this crate doesn't have a link to libcore, as that's needed for the trace function"?

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: Turns out it's only used once, so I think we should indeed inline it.

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: Good idea.

src/librustc_mir/transform/add_yk_swt_calls.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    // We can't add calls to promoted items.
+    if let Some(_) = src.promoted {
+        return false;
+    }
+
+    // We can't add calls to constant functions.
+    let node_id = tcx.hir.as_local_node_id(src.def_id)
+        .expect("Failed to get node id");
+    if let Some(fn_like) = FnLikeNode::from_node(tcx.hir.get(node_id)) {
```
ltratt, line 181: This can be simplified to something like:

```rust
if let Some(...) = ... {
    fn_like.constness() == hir::Constness::Const
} else {
    false
}
```

src/librustc_mir/transform/add_yk_swt_calls.rs
ltratt: Ah, so maybe you just need to say something like "Collectively, crate_hash, def_id, and block_index uniquely identify a MIR block"?

src/librustc_mir/transform/add_yk_swt_calls.rs
ltratt: Right, so it might be worth pulling these comments out to be something more like:

```
// In order to perform software tracing, we replace each MIR block with a new block
// that first records the block as having been traced and then calls the original block.
```

then I think the two variables are fairly obvious.

src/librustc_mir/transform/add_yk_swt_calls.rs
ltratt: So is it just one new decl or ...?

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: > Should this be

It should be clear from the name of the function that true means we should.

> annotate vs. should_be_traced

We are annotating blocks with calls, right?

I'm not sure about `should_be_traced`. How about `should_transform`?

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: OK, will review these.

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: Ah yes.

src/librustc_mir/transform/add_yk_swt_calls.rs
ltratt: You'd be amazed at how many functions that return bool aren't clear about what true/false means, so I find it best to be explicit. It's a bit like axes on graphs: often stating "higher is better" (or "lower is better") massively helps readers.

I'm actually wondering if we've got this function the wrong way around (so to speak). Maybe instead of "this can be annotated/transformed", it would be clearer to have the function be "can't be annotated/transformed"? Then the function might be called something like `is_untraceable`.

ltratt:
> Also, would I be an idiot if I couldn't see either being defined in 74a44dc? I think that commit is in the wrong place in the list?

Ah, I assume this is the github ordering problem you alluded to earlier? When we come to rebasing things later, it might be worth changing the date stamps on commits to make this a bit clearer?

vext01:
> Can you explain no_trace vs yk_swt_rec_loc?

Do you mean, why do we apply `no_trace` to the trace func?

> Also, would I be an idiot if I couldn't see either being defined in 74a44dc? I think that commit is in the wrong place in the list?

Github got the order of the commits wrong. See this comment:
https://github.com/softdevteam/ykrustc/pull/2#issuecomment-450656812

ltratt:
> Do you mean, why do we apply no_trace to the trace func?

No, I mean I'm not sure why we sometimes use one and sometimes the other.

vext01:
> No, I mean I'm not sure why we sometimes use one and sometimes the other.

`no_trace` is an attribute which can be applied at the item- or crate-level to suppress the MIR transform for that item or crate.

`yk_swt_rec_loc` is the function that the MIR pass adds calls to as long as the containing function or crate was not `no_trace`.

Makes sense?

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: Works for me.

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: OK, but I would like to try and capture exactly what happens in terms of the CFG too. I'll propose something that should keep us both happy.

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: Since we add one call for each block in the MIR, we add one new local decl per-block.

src/librustc_mir/transform/add_yk_swt_calls.rs
ltratt: It might not hurt to have an example (maybe -- god help us all! -- in ASCII art) too?

src/librustc_mir/transform/add_yk_swt_calls.rs
ltratt: Ah, right, so that's the crucial bit of missing information! That should go into the uber-comment I suggested a few lines above.

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: > Maybe instead of "this can be annotated/transformed", it would be clearer to have the function be "can't be annotated/transformed"

I'm not too fussed. Happy to give that a go.

On a side note, we will have to come back later and make a record (in the binary somewhere) of what was not traceable. We will need this at runtime. So perhaps your suggestion is a good one.

vext01:
>When we come to rebasing things later, it might be worth changing the date stamps on commits to make this a bit clearer?

We can do if you want. I don't think it matters too much though. It won't be the first or last time that someone re-orders commits in a re-base.

ltratt:
> We can do if you want. I don't think it matters too much though. It won't be the first or last time that someone re-orders commits in a re-base.

Point taken. And, IMHO, it is a github bug not to show them in parent/child order.

vext01:
> IMHO, it is a github bug not to show them in parent/child order.

I tend to agree.

Just to prove that the commits are ordered right:

```
$ git log
commit 74a44dc00adb50ad16b5d6df944e721531fc80ce
Author: Edd Barrett <vext01@gmail.com>
Date:   Mon Dec 31 14:52:13 2018 +0000

    Revise tests in light of Yorick software tracing MIR pass.

commit 3504c9c2c1c740a660a2120079c39aaa5defa1f4
Author: Edd Barrett <vext01@gmail.com>
Date:   Mon Dec 31 14:54:03 2018 +0000

    Add the Yorick software tracing MIR pass and associated functionality.

    This change:

     - Adds a MIR pass which adds, for each MIR block, a call to a tracing
       (language item) function. The arguments to this function uniquely
       identify a location in the MIR. A collection of such locations will
       form a trace.

     - Adds a "no_trace" attribute, which informs the compiler to not apply
       the above MIR pass to an item or crate. This is needed for tests, but
       also to prevent infinite recursion on the tracing function itself.

commit 7df215dfbe530ae9a5a3e622360e0b9b7ee138f0
Author: Edd Barrett <vext01@gmail.com>
Date:   Mon Dec 31 14:52:54 2018 +0000

    Make CI build stricter.

    This helps to find bugs earlier.

commit a9abb9a59c2ab2c7935582e9b3dade9c76571fca
Merge: 289ad6e992 98b0860a5c
Author: bors[bot] <bors[bot]@users.noreply.github.com>
Date:   Thu Nov 29 15:23:46 2018 +0000

    Merge #1
...
```

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: I was going to suggest ASCII art, but thought better of it! Let's give it a shot.

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: The `no_trace` annotations are not here, they are used in tests, and on the tracing function and its wrapper.

ltratt: Huh?

vext01:
How's this? I think it's an improvement :)

src/librustc_mir/transform/add_yk_swt_calls.rs
```diff
@@ -31,33 +51,24 @@ impl MirPass for AddYkSWTCalls {
...
-        // Types.
         let unit_ty = tcx.mk_unit();
         let u32_ty = tcx.types.u32;
         let u64_ty = tcx.types.u64;
 
-        // Each block is replaced by a new block whose terminator calls the recorder function.
-        let mut replace_blocks = Vec::new();
-
-        // The original blocks are copied and the recorder function returns to a copy.
+        let mut replacement_blocks = Vec::new();
```
ltratt: The convention in this function seems to be `blk` for `blocks`? I'm certainly a fan of the shorter version! It might be nice to find a shorter, and perhaps more accurate version of "replacement". Something like "shadow" perhaps?

src/librustc_mir/transform/add_yk_swt_calls.rs
```diff
@@ -18,8 +18,28 @@ use rustc::hir;
...
+/// The transformation works by copying each block and replacing the original with a new block
+/// which calls the trace recorder function. The call then returns to the copy.
+///
+/// For example:
+///
+/// +----+   +----+           +-----+   +-----+   +----+   +-----+   +-----+   +----+
+/// | B0 |-->| B1 |  Becomes: | B0' |-->| Rec |-->| B2 |-->| B1' |-->| Rec |-->| B3 |
+/// +----+   +----+           +-----+   +-----+   +----+   +-----+   +-----+   +----+
+///
+/// Where:
```
ltratt, line 34: I broadly like this, but the naming is really confusing (B0 becomes B0, B0', and B2?!). How about something like: B0 is the original; S0 is the shadow version of B0 (i.e. the copy we've made); and C0 is the copied version of B0 with a different terminator?

[This also goes along with my comment below about "replacement", which isn't a very helpful term here, since it suggests that B0 is entirely replaced. "Shadow" is my 10-second-solution, but maybe we want something which suggests "a new inserted thing which calls the old one"?]

src/librustc_mir/transform/add_yk_swt_calls.rs
```diff
@@ -131,51 +131,54 @@ impl MirPass for AddYkSWTCalls {
     }
 }
 
-/// Given a `MirSource`, decides if we should annotate the correpsonding MIR.
-fn should_annotate(tcx: TyCtxt<'a, 'tcx, 'tcx>, src: MirSource) -> bool {
-    // Never annotate any MIR-like thing marked `#[no_trace]` or `#[naked]`. The trace record and
-    // wrapper are also marked `#[no_trace]` to prevent infinite recursion.
+/// Given a `MirSource`, decides if it is possible for us to trace (and thus whether we should
+/// transform) the MIR. Returns `true` if we cannot trace, otherwise `false`.
+fn is_untraceable(tcx: TyCtxt<'a, 'tcx, 'tcx>, src: MirSource) -> bool {
```
ltratt, line 136: How about "Given a `MirSource`, return `false` if it should not be traced, or `true` otherwise"? The bit about transforming is, IMHO, irrelevant here: what the caller of this function chooses to do with the information is up to them.

src/librustc_mir/transform/add_yk_swt_calls.rs
```diff
@@ -131,51 +131,54 @@ impl MirPass for AddYkSWTCalls {
...
+    // Similar to `#[no_trace]`, don't transform anything inside a crate marked `#![no_trace]`.
     for attr in tcx.hir.krate_attrs() {
         if attr.check_name("no_trace") {
-            return false;
+            return true;
         }
     }
 
-    // We can't call the software tracing function if there is no libcore.
+    // We can't call the software tracing function if there is no libcore because that's where the
```
ltratt: When we say "no libcore" do we mean "if this crate doesn't have a link to libcore" or ...?

src/librustc_mir/transform/add_yk_swt_calls.rs
```diff
@@ -131,51 +131,54 @@ impl MirPass for AddYkSWTCalls {
...
+    // entry point to the trace recorder function lives.
     if attr::contains_name(tcx.hir.krate_attrs(), "no_core") {
-        return false;
+        return true;
     }
 
     // The libcompiler_builtins crate is special and we can't annotate it.
```
ltratt: s/can't/mustn't/? Also *why* mustn't we annotate it?

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: I'm OK with B -> S, but I'd like it if the indices matched the indices in the backing vector. I'm trying to convey the fact that we have to manipulate the MIR in the context of a list, and new blocks pushed will go on the end. So `Ci` I'm not so keen on.

ltratt: The structure of the CFG makes the ordering of the blocks completely clear doesn't it?

On the matter of the name, maybe we could call the new blocks "record blocks"? That gets their intent across quite clearly I think.

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: If this crate doesn't depend upon libcore, then we can't call something (the tracer func) in libcore.

ltratt: Aha, so I think "depend upon" is the key missing phrase then?

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: Hrm, something is only untraceable if it's not possible to apply the transform. i.e. "untraceable" here is defined in terms of the transform. The concepts are tightly coupled. Unless you feel strongly, I'd rather leave it in.

ltratt: You said above that you'll also need to call this function in contexts where you're not transforming anything, but simply want to know if it can/can't be traced. I know that "can't be traced" also means "can't be transformed", but this function is really only concerned about the former, isn't it?

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: To be honest, the compiler_builtins crate is a bit of a mystery to me. All I know is that strange things happen if you do transform it.

Ah: s/annotate/transform/ too.

If you like, we can remove this check, run bors, and observe the error message for clues as to what to put in this comment? I don't recall exactly what happens off the top of my head.

ltratt: It might be a good idea. If we've already forgotten, someone in the future's going to worry about it too. Since this is such a critical part of the whole infrastructure, I suspect our future selves will be grateful for such hints :)

vext01: Okie dokie! Here goes then.

vext01:
bors try

stand by for compilation error :)

vext01:
The relevant part of the log is:
```
undefined reference to `core::yk_swt::yk_swt_rec_loc_wrap'
```

I don't understand why this is the case, but since libcompiler_builtins is a load of wrapped C and asm code which we can't transform anyway, I'm still in favour of skipping this crate.

We can add a comment like:
```
/// Transforming libcompiler_builtins gives undefined references (linker magic?), but
/// even if it didn't the code within is just a load of wrapped C and asm code which we
/// can't transform anyway.
```

Thoughts?

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: This was in response to:
> I can't see anything in here that marks anything as no_trace?

ltratt:
More-or-less fine with me, but why don't we tell people that it's "core::yk_swt::yk_swt_rec_loc_wrap" that gives an error, just in case that rings bells with them?

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: > The structure of the CFG makes the ordering of the blocks completely clear doesn't it?

In the ASCII art diagram, the blocks are not shown in order they appear in the vector, which is why I named the blocks with their indices.

> On the matter of the name, maybe we could call the new blocks "record blocks"?

I'm not keen on "record blocks", because IDs that are recorded into the trace are in fact the shadows. How about `copy_blks`?

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: > You said above that you'll also need to call this function in contexts where you're not transforming anything, but simply want to know if it can/can't be traced

Maybe there's confusion with the `no_trace` attribute? The `is_untraceable` function has only one call site.

vext01:
OK, there's a few more fixes.

There's a few comments awaiting.

src/librustc_mir/transform/add_yk_swt_calls.rs
ltratt: After a quick conversation we're thinking that "shadow blocks" and "user blocks" might be appropriate terms.

src/librustc_mir/transform/add_yk_swt_calls.rs
ltratt: OK.

vext01:
That's the last fix in. Happy for me to squash?

ltratt:
Please squash.

vext01:
splat.

ltratt:
I don't think you've pushed anything?

vext01:
Oops! Try now.

PR #3, opened 2019-01-10
Preliminary trace recording.

vext01:
This shows that we can store MIR locations into a thread local vector at runtime.

It's a quick and dirty implementation that disables tracing for a bunch of crates that ideally we probably do want to trace at some point. As discussed on IRC, we are going to live with this compromise until a later date, where we might (e.g.) implement the trace recorder in another language that the MIR pass can't see.

I've added some tests of our own too.

I've run the smoke test subset of the rustc tests at stage one and fixed a few failures. Let's see what happens in a full run.

Any comments in the meantime?

### DIFF
--- a/src/librustc_mir/transform/add_yk_swt_calls.rs
+++ b/src/librustc_mir/transform/add_yk_swt_calls.rs
@@ -18,6 +18,11 @@ use rustc::hir;
 use rustc::hir::def_id::{DefIndex, LOCAL_CRATE};
 use rustc::hir::map::blocks::FnLikeNode;
 
+// Crates the trace recorder depends upon.
+// This is effectively libstd and its dependencies (from its Cargo manifest).
+static RECORDER_DEPS: &'static [&'static str] = &["std", "alloc", "panic_unwind",
+    "panic_abort", "core", "libc", "compiler_builtins", "profiler_builtins", "unwind"];
+
 /// A MIR transformation that, for each basic block, inserts a call to the software trace recorder.
 /// The arguments to the calls (crate hash, DefId and block index) identify the position to be
 /// inserted into a trace.
@@ -167,6 +172,12 @@ fn is_untraceable(tcx: TyCtxt<'a, 'tcx, 'tcx>, src: MirSource) -> bool {
     // recorder wrapper `core::yk_swt::yk_swt_rec_loc_wrap`. It's not worth investigating, as this
     // crate only contains wrapped C and ASM code that we can't transform anyway.
     if tcx.is_compiler_builtins(LOCAL_CRATE) {
+        return true;
+    }
+
+    // FIXME Anything which the trace recorder depends upon cannot be traced or we will get
+    // infinite recursion at runtime..
+    if RECORDER_DEPS.contains(&&*tcx.crate_name(LOCAL_CRATE).as_str()) {
         return true;
     }
 
--- a/src/librustc_mir/transform/add_yk_swt_calls.rs
+++ b/src/librustc_mir/transform/add_yk_swt_calls.rs
@@ -18,10 +18,10 @@ use rustc::hir;
 use rustc::hir::def_id::{DefIndex, LOCAL_CRATE};
 use rustc::hir::map::blocks::FnLikeNode;
 
-// Crates the trace recorder depends upon.
-// This is effectively libstd and its dependencies (from its Cargo manifest).
-static RECORDER_DEPS: &'static [&'static str] = &["std", "alloc", "panic_unwind",
-    "panic_abort", "core", "libc", "compiler_builtins", "profiler_builtins", "unwind"];
+// These crates are not transformed by this MIR pass.
+// FIXME: Currently libstd and its deps are blacklisted. In the long run, we don't want this.
+static BLACKLISTED_CRATES: &'static [&'static str] = &["std", "alloc", "panic_unwind",
+    "panic_abort", "core", "libc", "profiler_builtins", "unwind"];
 
 /// A MIR transformation that, for each basic block, inserts a call to the software trace recorder.
 /// The arguments to the calls (crate hash, DefId and block index) identify the position to be
@@ -175,9 +175,8 @@ fn is_untraceable(tcx: TyCtxt<'a, 'tcx, 'tcx>, src: MirSource) -> bool {
         return true;
     }
 
-    // FIXME Anything which the trace recorder depends upon cannot be traced or we will get
-    // infinite recursion at runtime..
-    if RECORDER_DEPS.contains(&&*tcx.crate_name(LOCAL_CRATE).as_str()) {
+    // Don't transform any crate which is blacklisted. See the comment against BLACKLISTED_CRATES.
+    if BLACKLISTED_CRATES.contains(&&*tcx.crate_name(LOCAL_CRATE).as_str()) {
         return true;
     }
 
--- a/src/libstd/yk_swt.rs
+++ b/src/libstd/yk_swt.rs
@@ -7,6 +7,45 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+use ::cell::RefCell;
+use ::fmt;
+
+#[allow(missing_docs)]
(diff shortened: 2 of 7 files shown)


vext01:
bors try

src/librustc_mir/transform/add_yk_swt_calls.rs
```diff
@@ -18,6 +18,11 @@ use rustc::hir;
 use rustc::hir::def_id::{DefIndex, LOCAL_CRATE};
 use rustc::hir::map::blocks::FnLikeNode;
 
+// Crates the trace recorder depends upon.
+// This is effectively libstd and its dependencies (from its Cargo manifest).
+static RECORDER_DEPS: &'static [&'static str] = &["std", "alloc", "panic_unwind",
```
ltratt: I think it might be clearer to call this something like `CRATES_TRACE_BLACKLIST`: in a sense, the fact that the crate is a dependency is less relevant than that we're going to avoid tracing it?

src/libstd/yk_swt.rs
```diff
@@ -7,6 +7,45 @@
...
+
+#[allow(missing_docs)]
+/// A block location in the Rust MIR.
+pub struct MirLoc {
+    pub crate_hash: u64,
+    pub def_idx: u32,
+    pub bb_idx: u32,
+}
+
+impl fmt::Debug for MirLoc {
```
ltratt, line 21: Is it worth this manual code vs. `#[derive(Debug)]`?

src/librustc_mir/transform/add_yk_swt_calls.rs
vext01: OK.

src/libstd/yk_swt.rs
vext01: The reason I did a custom one was because the derived one is quite long, with the names of the fields etc.

I wanted a shorter representation.

src/libstd/yk_swt.rs
```diff
@@ -7,6 +7,45 @@
...
+        let mut trace_o = rc.borrow_mut();
+        match *trace_o {
+            Some(_) => panic!("tracing was already started for this thread!"),
+            None => *trace_o = Some(Vec::new()),
+        }
+    });
+}
+
+// FIXME Anything used in `rec_loc` below cannot itself be traced, or we get infinite recursion. To
+// work sround this, many crates are ignored by the software tracing MIR pass (see
```
ltratt: "Around".

vext01: oopsy!

src/libstd/yk_swt.rs
```diff
@@ -15,5 +54,23 @@
 #[cfg_attr(not(stage0), no_trace)]
 #[cfg(not(test))]
 fn rec_loc(crate_hash: u64, def_idx: u32, bb_idx: u32) {
-    // Not implemented.
+    TRACE.with(|rc| {
+        let mut trace_o = rc.borrow_mut();
+        match trace_o.as_mut() {
+            Some(trace) => trace.push(MirLoc{crate_hash, def_idx, bb_idx}),
+            None => (), // Tracing is disabled, do nothing.
```
ltratt: "Disabled" or "not active"?

vext01: Both I guess. How about:

"We are not currently tracing, do nothing"

src/libstd/yk_swt.rs
```diff
@@ -15,5 +54,23 @@
...
+        }
+    });
+}
+
+/// Stop tracing and return the trace.
+#[cfg_attr(not(stage0), no_trace)]
+pub fn stop_tracing() -> Vec<MirLoc> {
+    TRACE.with(|rc| {
+        let trace_o = rc.borrow_mut().take();
```
ltratt: Can't this be something like:

```rust
rc.borrow_mut().take.expect("tracing not started on this thread")
```

?

src/test/run-pass/yk_swt/collect-trace.rs
```diff
@@ -0,0 +1,28 @@
...
+pub fn main() {
+    start_tracing();
+    let _ = work();
+    let trace = stop_tracing();
+    assert!(!trace.is_empty());
+}
+
+#[inline(never)]
+fn work() -> u64{
+    let mut res = 47;
```
ltratt: Any reason for the unusual numbers?

src/libstd/yk_swt.rs
vext01: I think it could, yeah. Thanks.

src/test/run-pass/yk_swt/collect-trace.rs
vext01: Arbitrary.

vext01:
Thanks for the comments. I'm going to wait for buildbot before applying changes.

src/libstd/yk_swt.rs
ltratt: OK, but let's only do this sparingly, just because it's that bit more code to read and maintain.

src/libstd/yk_swt.rs
ltratt: Agreed.

src/test/run-pass/yk_swt/collect-trace.rs
ltratt: Can I suggest we choose less weird numbers? Otherwise it makes me think there's something specific about these (perhaps some threshold needs to be exceeded). I am a big fan of numbers < 10, or multiples of 10 ;)

ltratt:
You can, I think, push changes even before buildbot has finished without affecting it. But it's up to you -- I'm easy either way!

src/test/run-pass/yk_swt/collect-trace.rs
vext01: Can do, yeah.

vext01:
I configured it to run a build for each push. So perhaps I'll make the changes locally and wait for the push.

ltratt:
> I configured it to run a build for each push. So perhaps I'll make the changes locally and wait for the push.

Dumb question: what's the point of manually doing "try" then? I assumed we had to manually kick it to get it to do anything...

vext01:
You are right, I think I must be wrong.

src/libstd/yk_swt.rs
vext01: We are only resolving threads when I have pushed the fixes, right? This one was marked resolved.

vext01:
I'll start looking at the test failures first thing tomorrow. Cheers.

src/libstd/yk_swt.rs
ltratt: We had resolved the discussion (and I trust you to do what you'd suggested you would do). No point having things hanging around in such a case I think?

src/librustc_mir/transform/add_yk_swt_calls.rs
```diff
@@ -175,9 +175,8 @@ fn is_untraceable(tcx: TyCtxt<'a, 'tcx, 'tcx>, src: MirSource) -> bool {
         return true;
     }
 
-    // FIXME Anything which the trace recorder depends upon cannot be traced or we will get
-    // infinite recursion at runtime..
-    if RECORDER_DEPS.contains(&&*tcx.crate_name(LOCAL_CRATE).as_str()) {
+    // Don't transform any crate which is blacklisted. See the comment against BLACKLISTED_CRATES.
```
ltratt, line 178: I don't think we need this comment (it just says what the line of code below does, and the line of code is very clear IMHO).

src/libstd/yk_swt.rs
vext01: Ah I see. This isn't quite what I thought. The problem with this strategy is that I have to go unfolding threads to see if I've addressed them or not (in code). I thought you would only resolve them when you saw the commit that addresses them.

But that might also be sub-optimal?

ltratt:
Is this one still moving?

vext01:
This was rendered dead by #4

PR #8, opened 2019-03-12
Sync upstream to get proc-macro goodness.

vext01:
This syncs us with upstream so we can have the proc-macro stuff needed for in-compiler serde.

Sorry it took so long, the merge was non-trivial, and I messed it up first time.


vext01:
bors try

PR #10, opened 2019-03-15
Start converting MIR to SSA TIR.

vext01:
This change:

 - Starts adding support for statements to TIR.
   Only simple assignments are translated. Everything else is made into
   an "unimplemented" statement. The plan is to keep going until the BF
   interpreter has none of these.

 - Does the first part of the SSA conversion, namely:
   * Computes the dominance frontier.
   * Inserts PHI nodes in the right place and for the correct TIR variables.

So what we get is a CFG with PHIs, but which is not yet in SSA. The
remaining step is to walk the dominator tree and assign each definition
of a variable with a new SSA variable name. That's yet another algorithm
from the Appel book. To keep PRs manageable, I think we should do that
as a separate PR.

Manual inspection of CFG graphs with assignments and PHIs shown
look good. Here's a couple of examples:

## example 1

image

## example 2

image

I'd like to add some tests to this code to try and weed out some early
bugs (I'm sure there will be some). This is likely to take me a little
while as I want to inspect how upstream does this kind of testing. I
want to avoid manually constructing MIRs and TIRs by hand if I can help
it.

Question: Currently PHI nodes are of arbitrary arity to make them easier
to read during development. We talked about having only binary PHIs. Is
that still what we want?

[There is a local ykpack Cargo path until we decide what to do with
PHIs arity]

Cheers.


src/librustc_yk_sections/Cargo.toml
```diff
@@ -12,5 +12,8 @@ test = false
 [dependencies]
 rustc = { path = "../librustc" }
 rustc_yk_link = { path = "../librustc_yk_link" }
-ykpack = { git = "https://github.com/softdevteam/ykpack" }
+#ykpack = { git = "https://github.com/softdevteam/ykpack" }
```
vext01: > There is a local ykpack Cargo path until we decide what to do with
PHIs arity

So ignore this for now. I'll have a PR for ykpack once we've talked about PHI arity.

vext01:
By the way, the variable names in the alrogithmy parts of the code are named the same as in the book. So if you are wondering why the names are so obviously not "edd-like", that's why :)

ltratt:
I have no strong opinions on arity: clearly binary phi nodes are expressive enough to implement everything else in, and for the (probably) common case they're the most efficient way of doing it. I'd probably be inclined to start with binary phi nodes and change later if that turns out to be unduly awkward.

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -18,20 +16,95 @@ use rustc::ty::TyCtxt;
...
 
+type TirVarIndex = u32;
+
+/// A conversion context holds the state needed to perform the conversion to (pre-SSA) TIR.
+struct ConvCx<'a, 'tcx, 'gcx> {
+    /// The compiler's god struct. Needed for queries etc.
+    tcx: &'a TyCtxt<'a, 'tcx, 'gcx>,
+    /// The definition sites of TIR variable in terms of basic blocks.
+    def_sites: RefCell<Vec<Vec<BasicBlock>>>,
+    /// The next TIR variable index to be issued.
```
ltratt: Huh? Is this a fancy-pants yet vague way of saying "This is a monotonically increasing integer which is used to give Tir nodes a unique ID"?

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -18,20 +16,95 @@ use rustc::ty::TyCtxt;
...
+
+        Self {
+            tcx,
+            def_sites: RefCell::new(Vec::new()),
+            next_tir_var: Cell::new(0),
+            var_map: RefCell::new(var_map),
+        }
+    }
+
+    /// Issues a fresh TIR variable index.
```
ltratt: Something like "Returns a guaranteed unique TIR variable index"? "Fresh" isn't exactly wrong, but I've only ever seen it used to refer to variable *names*.

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -18,20 +16,95 @@ use rustc::ty::TyCtxt;
...
+    fn next_tir_var(&self) -> TirVarIndex {
+        let var_idx = self.next_tir_var.get();
+        self.next_tir_var.set(var_idx + 1);
+        var_idx
+    }
+
+    /// Get the TIR variable corresponding with the MIR variable, creating a fresh one if needed.
```
ltratt: Should this function be called `mir_var_to_tir` or similar?

src/librustc_yk_sections/mir_cfg.rs
vext01: Yes, but it's current value is the one which will be given out next.

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -18,20 +16,95 @@ use rustc::ty::TyCtxt;
...
+    fn get_tir_var(&self, local: Local) -> TirVarIndex {
+        let local_u32 = local.as_u32();
+        let mut var_map = self.var_map.borrow_mut();
+
+        // Resize the backing Vec if necessary.
+        if var_map.len() <= usize::try_from(local_u32).unwrap() {
```
ltratt: I would prefer to do this as:

```rust
debug_assert!(usize::max_value >= u32::max_value());
if var_map.len <= usize::from(local.as_u32()) { ... }
```

because the `try_from` is never really necessary (we're never going to run on a 16 bit machine) so we might as well avoid the performance hit.

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -18,20 +16,95 @@ use rustc::ty::TyCtxt;
...
+            var_map.resize(usize::try_from(local_u32 + 1).unwrap(), None);
```
ltratt: The `+` here needs to be `checked_add` because we could overflow.

src/librustc_yk_sections/mir_cfg.rs
vext01: [I'm likely to abstract this mechanism in the next PR, as the last stage requires us to re-number all of our variables in a similar fashion. The reason there is a mapping in the conversion is because I don't think we will get far with a one-to-one mapping from MIR to TIR variables, unless we also model projections in the same way as Rust does.]

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -18,20 +16,95 @@ use rustc::ty::TyCtxt;
...
+        }
+
+        let v = match var_map[local] {
```
ltratt: This looks like `unwrap_or_else` in disguise?

ltratt: i.e. this could be `var_map[local].unwrap_or_else(|| ...)` without the `let v = ` or `match` etc.

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -18,20 +16,95 @@ use rustc::ty::TyCtxt;
...
+
+    fn def_sites(self) -> Vec<Vec<BasicBlock>> {
+        self.def_sites.into_inner()
+    }
+
+    /// Add `bb` as a definition site of the TIR variable `var`.
+    fn push_def_site(&self, bb: BasicBlock, var: TirVarIndex) {
+        let mut sites = self.def_sites.borrow_mut();
+        if sites.len() <= usize::try_from(var).unwrap() {
+            sites.resize((var + 1) as usize, Vec::new());
```
ltratt: `+`  probably needs to be `checked_add`.

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -50,8 +123,21 @@ pub fn generate_tir<'a, 'tcx, 'gcx>(
...
+
+            // Get an initial TIR (not yet in SSA form).
+            let pre_ssa_pack = (&ccx, def_id, tcx.optimized_mir(*def_id)).to_pack();
+
+            // Add PHI nodes.
+            let phied_pack = PhiInserter::new(mir, pre_ssa_pack, ccx.def_sites()).pack();
+
+            // FIXME - rename variables with fresh SSA names.
+
+            // Put the finalised TIR to disk.
```
ltratt: This comment is redundant.

src/librustc_yk_sections/mir_cfg.rs
vext01: It could be... I don't think it improves the meaning much though!

"give me an index for this MIR variable, make a new index if we havne't seen it yet".

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -64,43 +150,192 @@ pub fn generate_tir<'a, 'tcx, 'gcx>(
     Ok(ret)
 }
 
+/// Lazy computation of dominance frontiers.
+///
+/// See Page 404 of the 2nd edition of the Appel book mention above.
```
ltratt: "mentioned". Can also simplify "Page 404" to "p404" if you want.

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -64,43 +150,192 @@ pub fn generate_tir<'a, 'tcx, 'gcx>(
...
+    /// The dominators of the above MIR.
+    doms: Dominators<BasicBlock>,
+    /// The dominance frontiers. Lazily computed. None means as-yet uncomputed.
+    df: IndexVec<BasicBlock, Option<Vec<BasicBlock>>>,
+}
+
+impl<'a, 'tcx> DominatorFrontiers<'a, 'tcx> {
+    fn new(mir: &'a Mir<'tcx>) -> Self {
+        let num_blks = mir.basic_blocks().len();
+        let mut df = IndexVec::with_capacity(num_blks);
```
ltratt: Does `IndexVec` have a `from_elem` function? If so, it'll probably execute faster than this + the for loop.

src/librustc_yk_sections/mir_cfg.rs
vext01: Ah, I was meant to ask about try_from in the PR description.

We have to convert in both directions. From u32 to usize and back.

The u32 to usize direction can be solved as you say. What about the inverse?

src/librustc_yk_sections/mir_cfg.rs
vext01: Agreed

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -64,43 +150,192 @@ pub fn generate_tir<'a, 'tcx, 'gcx>(
...
+                }
+            }
+
+            // The second stage of the algorithm needs the children nodes in the dominator tree.
+            let mut children = Vec::new();
+            for (b, _) in self.mir.basic_blocks().iter_enumerated() {
+                let b = BasicBlock::from_u32(b.as_u32());
+                if self.doms.is_dominated_by(b, n) && b != n {
+                    children.push(b);
+                    // Force the frontier of `b` into `self.df`. Doing this here avoids a
```
ltratt: Not sure I understand the comment or why `get` does anything.

src/librustc_yk_sections/mir_cfg.rs
vext01: I'll try. This part was hard to get pass the borrow checker.

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -64,43 +150,192 @@ pub fn generate_tir<'a, 'tcx, 'gcx>(
...
+
+        // We first need a mapping from block to the variables it defines. Appel calls this
+        // `A_{orig}`. We can derive this from our definition sites.
+        let (a_orig, num_blks) = {
+            let ykpack::Pack::Mir(ykpack::Mir{ref blocks, ..}) = self.pack;
+            let mut a_orig: IndexVec<BasicBlock, Vec<TirVarIndex>> = IndexVec::with_capacity(blocks.len());
+            a_orig.resize(blocks.len(), Vec::default());
+            for (a, def_blks) in self.def_sites.iter().enumerate() {
+                for bb in def_blks {
+                    a_orig[*bb].push(u32::try_from(a).unwrap())
```
ltratt: What type does `a` have? Does this need to be `try_from`?

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -64,43 +150,192 @@ pub fn generate_tir<'a, 'tcx, 'gcx>(
...
+
+            (a_orig, blocks.len())
+        };
+
+        let mut a_phi: Vec<Vec<TirVarIndex>> = Vec::with_capacity(num_blks);
+        a_phi.resize(num_blks, Vec::new());
+        for (a, def_sites) in self.def_sites.iter().enumerate() {
+            let mut w = def_sites.clone();
+            while w.len() > 0 {
+                let n = w.pop().unwrap();
```
ltratt: The intermediate variable seems of little use?

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -64,43 +150,192 @@ pub fn generate_tir<'a, 'tcx, 'gcx>(
...
+                }
+            }
+
+            (a_orig, blocks.len())
+        };
+
+        let mut a_phi: Vec<Vec<TirVarIndex>> = Vec::with_capacity(num_blks);
+        a_phi.resize(num_blks, Vec::new());
+        for (a, def_sites) in self.def_sites.iter().enumerate() {
+            let mut w = def_sites.clone();
```
ltratt: Hang on, below we push into `w`, which is a clone of `def_sites`, so I think this loop has no effect on `def_sites` or ...?

ltratt:
OK, so we definitely need tests for this, even if they're very basic ones. In my experience, these sorts of algorithms are too easy to get subtly wrong (or, at least, I get them wrong often!).

src/librustc_yk_sections/mir_cfg.rs
ltratt: It doesn't matter if this is the current or next index: it's the monotonically increasing bit which is important.

src/librustc_yk_sections/mir_cfg.rs
ltratt: In general, Rust has a dislike of `get` functions, and I've come to think this is generally a good idea.

src/librustc_yk_sections/mir_cfg.rs
ltratt: In general usize to u32 is clearly unsafe on a 64 bit machine. *But* if we know that we only ever wrote out 32 bit numbers, the inverse is safe. It's fine to do something like this:

```rust
// Since we only ever issue numbers which fit into 32 bits, this conversion
// is guaranteed to be safe.
let x: usize = ...;
let y: u32: x as u32;
```

But each case must be clearly documented.

src/librustc_yk_sections/mir_cfg.rs
vext01: Oh, this part of the code was horrendous to get into rust.

TLDR: `get()` has side effects.

Long answer:  if were were literally translating the algorithm from the book,  we'd have done the get in the next chunk of the algorithm, but it's impossible to get past the borrow checker. Get mutates an internal cache inside the dominator frontier struct. By mutating now, we can ensure the entry exists and in the next chunk do a regular index operation without any mutation.

The cache exists because the dominator frontier algorithm is like a big pot of simultaneous equations and without a cache we'd compute the same things over and over. It's a bit like memoising fib.

Hard to explain. Does it make sense?

src/librustc_yk_sections/mir_cfg.rs
vext01: A is a `usize` here as it comes from `enumerate`.

src/librustc_yk_sections/mir_cfg.rs
vext01: how about just `tir_var()` then?

src/librustc_yk_sections/mir_cfg.rs
ltratt: If `get` has side effects, it seems like a weird name for that function!

src/librustc_yk_sections/mir_cfg.rs
vext01: OK, seems fine by me.

[I think Rust really makes this harder than it needs to be :( ]

ltratt: Don't start me on my Rust-integer-conversions-aren't-great rant!

src/librustc_yk_sections/mir_cfg.rs
vext01: Hrm. Interesting. At the time of writing I think that was intentional, but now I look again, the def site for any given variable is not used subsequently, so I *think* we can operate on the list in-place.

image

src/librustc_yk_sections/mir_cfg.rs
vext01: Solely to pair with the original algorithm:

image

src/librustc_yk_sections/mir_cfg.rs
ltratt: OK.

src/librustc_yk_sections/mir_cfg.rs
ltratt: I think that text needs to be bigger ;)

Anyway, fair enough.

src/librustc_yk_sections/mir_cfg.rs
vext01: It could have had to do with the borrow checker ;) I'll see what I can do!

vext01:
> OK, so we definitely need tests for this

Yes, totally agreed!

Especially since it worked too quickly. 2nd or 3rd try after getting it through the borrow checker! I fear things which work close to first time...

vext01:
> I'd probably be inclined to start with binary phi nodes and change later if that turns out to be unduly awkward.

It's actually easier for me to understand n-ary PHI nodes.

Imagine what this would look like once you've unrolled all of the n-ary PHIs:

image

So unless you object, I'd rather start with n-ary and change to binary if/when we have a good reason to do so. Would that be ok?

ltratt:
What do other people do?

vext01:
It looks like LLVM uses n-ary PHI nodes:
https://llvm.org/docs/LangRef.html#phi-instruction

ltratt:
OK.

src/librustc_yk_sections/mir_cfg.rs
vext01: We can call it whatever we like. In the book it's a regular hash-table operation.

How about `frontier()` because that's what it returns.

ltratt: That seems like a better name.

src/librustc_yk_sections/mir_cfg.rs
vext01: Yes, it does. Much neater.

src/librustc_yk_sections/mir_cfg.rs
vext01: I wonder if we actually want to check the size of the integer in bytes, not the maximum values:

```
77 |         debug_assert!(usize::max_value() >= u32::max_value());
   |                                             ^^^^^^^^^^^^^^^^ expected usize, found u32
```

If we don't know which is larger out of usize and u32, we can't insert a correct cast on either the left or right hand side of the assertion, right?

So should we use?:
```
use std::mem::size_of;
debug_assert!(size_of(usize) >= size_of(u32));
```

I think that's robust as long as we are careful not to mix signedness.

ltratt: In practise these are clearly equivalent *but* in theory the size of a thing doesn't guarantee the range of values it can contain (Rust makes no guarantees about bit representations IIRC), so `max_value` is perhaps marginally safer in an angels-on-the-head-of-a-pin sense.

vext01: So I think you are suggesting that we assume usize is wider than u32 and we do:
```
debug_assert!(usize::max_value() >= usize::from(u32::max_value()));
```

But if we are assuming exactly what we are asserting, then why have the assertion at all?

Maybe I'm missing something?

Worded differently, how would you fix the above compilation error in a meaningful way?

ltratt: The point of the assertion is that if, somehow, this gets compiled on a platform where the assertion isn't true, it'll leave behind a panic. Thus the assertion implicitly guarantees that our assumption doesn't silently cause problems.

vext01: Yes, I understand that :)

I think I'm failing to communicate my point. Let me try again.

You suggested an assertion like this:
```
debug_assert!(usize::max_value() >= u32::max_value());
```

This can't work because the max value on the left hand side of the `>=` evaluates to a `usize` and the max value on the right is a `u32`. So we cannot directly compare them.

We'd need a safe cast on either the left or right, but I can't see how if you don't know which of a `usize` or a `u32` is bigger.

We can assume a `usize` can express all of the values of a `u32` and do:
```
debug_assert!(usize::max_value() >= usize::from(u32::max_value()));
```

But if we assume a `usize` can express all of the values of a `u32`, then the assertion is a tautology from the start.

Do you follow?

ltratt: Ah, I'm with you! Yes, you're right. In fact, my past self knew this too ;)

```
$ cd grmtools
$ rg size_of|grep assert
cfgrammar/src/lib/idxnewtype.rs:                debug_assert!(size_of::<usize>() >= size_of::<T>());
cfgrammar/src/lib/idxnewtype.rs:                debug_assert!(size_of::<u32>() >= size_of::<T>());
lrpar/src/lib/lex.rs:        debug_assert!(size_of::<usize>() >= size_of::<u32>());
lrpar/src/lib/lex.rs:            debug_assert!(size_of::<usize>() >= size_of::<u32>());
lrtable/src/lib/mod.rs:        debug_assert!(size_of::<usize>() >= size_of::<StIdxStorageT>());
```

vext01: Cool!

vext01: By the way, in cases like:

```
debug_assert!(size_of::<usize>() >= size_of::<T>());
```

Might be worth double checking `T` must be unsigned?

ltratt: Yes, that is enforced through the `num_traits` crate.

vext01: Cool, one more thing, you suggested:
```
if var_map.len <= usize::from(local.as_u32()) { ... }
```

But I think it'll need to be as `as` conversion, due to:
```
79 |         if var_map.len() <= usize::from(local_u32).unwrap() {
   |                             ^^^^^^^^^^^ the trait `std::convert::From<u32>` is not implemented for `usize`
```

Is that OK?

ltratt: The assertion guarantees that the `as` is safe, right? But, in general, we should add a comment at each `as` site documenting why it's safe (e.g. "this is guaranteed by the execution in `f`, which must have been called before this function").

vext01: Agreed on both counts.

vext01:
With regards to tests, and further to our MM convo... (cc @jacob-hughes)

Having read how the black box tests work in rustc, I think what we can do is extend the `compiletest` tool to add our own suite kind.

Our suite would be quite similar to the `MirOpt` test suite:
https://github.com/rust-lang/rust/blob/0f88167f89fffe321590c5148f21b7d51d44388d/src/tools/compiletest/src/runtest.rs#L2864-L2877

So you can imagine that we would have a `run_tir_test()` which dumps the *TIR* (instead of the MIR) of the input and then does matching against it.

I think by doing it like this we can borrow a lot of existing infra. But we would need to write our own TIR -> text dumper.

I'm toying with the idea of doing tests as a fresh PR, as this one is already quite big.

Would that be OK?

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -61,34 +61,32 @@ impl<'a, 'tcx, 'gcx> ConvCx<'a, 'tcx, 'gcx> {
...
+    fn tir_var(&self, local: Local) -> TirVarIndex {
         let local_u32 = local.as_u32();
         let mut var_map = self.var_map.borrow_mut();
 
         // Resize the backing Vec if necessary.
-        if var_map.len() <= usize::try_from(local_u32).unwrap() {
-            var_map.resize(usize::try_from(local_u32 + 1).unwrap(), None);
+        // Vector indices are `usize`, but variable indices are `u32`, so converting from a
+        // variable index to a vector index is always safe if a `usize` can express all `u32`s.
+        debug_assert!(size_of::<usize>() >= size_of::<u32>());
```
ltratt: This should be an `assert`. If it's true, the optimiser will remove it, but it means that even people on release mode will find out if (somehow) our code will break.

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -61,34 +61,32 @@ impl<'a, 'tcx, 'gcx> ConvCx<'a, 'tcx, 'gcx> {
...
-                let var_idx = self.next_tir_var();
-                var_map[local] = Some(var_idx);
-                var_idx
-            }
-            Some(tvar) => tvar,
-        };
-
-        v
+        var_map[local].unwrap_or_else(|| {
+            // Make a new variable.
```
ltratt: I'm not sure this comment is really needed?

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -98,10 +96,16 @@ impl<'a, 'tcx, 'gcx> ConvCx<'a, 'tcx, 'gcx> {
     /// Add `bb` as a definition site of the TIR variable `var`.
     fn push_def_site(&self, bb: BasicBlock, var: TirVarIndex) {
         let mut sites = self.def_sites.borrow_mut();
-        if sites.len() <= usize::try_from(var).unwrap() {
-            sites.resize((var + 1) as usize, Vec::new());
+        // Vector indices are `usize`, but variable indices are `u32`, so converting from a
```
ltratt: I would probably delete this comment, and then put (above the `let var_usize` line) a short comment along the lines of "var is u32 which the earlier assert guarantees can fit into a usize".

src/librustc_yk_sections/mir_cfg.rs
vext01: Good point. Shall I change all of the numeric range debug assertions to full on assertions?

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -255,7 +256,9 @@ impl<'a, 'tcx> PhiInserter<'a, 'tcx> {
             a_orig.resize(blocks.len(), Vec::default());
             for (a, def_blks) in self.def_sites.iter().enumerate() {
                 for bb in def_blks {
-                    a_orig[*bb].push(u32::try_from(a).unwrap())
+                    // The `def_sites` vector is guaranteed to have no more entries than a `u32`
```
ltratt: We can shorten this to something like e.g. "def_sites is guaranteed to have at most u32::max_value() items".

src/librustc_yk_sections/mir_cfg.rs
ltratt: Yep.

src/librustc_yk_sections/mir_cfg.rs
vext01: Delete the assertion too, I think?

ltratt: Only if the assertion was 100% guaranteed to have been made my some other bit of code (and, if so, we should document it).

vext01: The assertion is identical to the earlier one you referenced.

But let me suggest we leave it in, since it protects from the other being removed later, and we've established that these assertions are pretty much free. What do you think?

ltratt: Yes, I know it's identical, but *is it guaranteed to have been called before this code is executed*?

vext01: Ah sorry. I see.

Yes, the variable will have to have been created prior to it's def site being pushed.

vext01:
I think that's all the comments addressed.

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -96,9 +95,7 @@ impl<'a, 'tcx, 'gcx> ConvCx<'a, 'tcx, 'gcx> {
     /// Add `bb` as a definition site of the TIR variable `var`.
     fn push_def_site(&self, bb: BasicBlock, var: TirVarIndex) {
         let mut sites = self.def_sites.borrow_mut();
-        // Vector indices are `usize`, but variable indices are `u32`, so converting from a
-        // variable index to a vector index is always safe if a `usize` can express all `u32`s.
-        debug_assert!(size_of::<usize>() >= size_of::<u32>());
+        // `var` is `u32`. The conversion up to `usize` was checked above in `tir_var()`.
```
ltratt: But is `tir_var` guaranteed to have been called before `push_def_site`? I think so, at which point something like "This conversion is safe because TirVarIndex was generated by tir_var which guarantees that a u32 can fit in a usize" is probably the right summary.

vext01: I've just double checked. We convert the "Place" first, which makes a TIR variable, and then after that we push a def site.

I'll update the comment.

ltratt: OK.

vext01:
OK, give that a go.

ltratt:
Please squash. I agree that we can do the tests in the next PR.

vext01:
There's the squash, but the tidy check is failing. One moment.

vext01:
Pretty sure you will be OK with these, but we have to follow protocol :)

Can I squash those?

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -400,11 +401,14 @@ impl<'tcx> ToPack<ykpack::Terminator> for (&ConvCx<'_, 'tcx, '_>, &Terminator<'t
 }
 
 /// BasicBlockData -> Pack
-impl<'tcx> ToPack<ykpack::BasicBlock> for (&ConvCx<'_, 'tcx, '_>, BasicBlock, &BasicBlockData<'tcx>) {
+impl<'tcx> ToPack<ykpack::BasicBlock> for
+    (&ConvCx<'_, 'tcx, '_>, BasicBlock, &BasicBlockData<'tcx>) {
+
```
ltratt: Random blank line?

vext01: It looks weird when a line continuation is butted up against a "real" indent.

ltratt: Put the curly on the next line then?

vext01:
Like this?

ltratt:
Yep! Please squash.

vext01:
Squashed and updated commit message.

ltratt:
@vext01 Not sure if you've seen the failure here?

vext01:
Yep. Working on it now.

I realised on the walk in that I'd forgotten to submit the ykpack changes!

vext01:
Not yet :)

vext01:
OK, there's three more changes. See commit messages.

One of them (the index type change) will need squashing in if you agree.

Sorry to sneak these in last minute.

ltratt:
Please squash.

vext01:
Squashed. Now we are ready for bors :)

vext01:
> invalid source: "git+https://github.com/softdevteam/ykpack#358a7403a5b8c03809f5c5b07933a6850060afa4"

I'm on it.

vext01:
This should do the trick. OK to squash again?

ltratt:
Please squash.

vext01:
Splat -- also squashed away "use upstream ykpack".

vext01:
Ran out of memory.

I think this is likely due to assertions and debuginfo.

In may 2018, people reported >4GiB memory usage (presumably on a normal non-debug build).

https://internals.rust-lang.org/t/rustc-build-memory-requirement/7457/10

Shall we up from 5GiB to say 8?

ltratt:
I think that's sensible. Please squash that in at the same time.

vext01:
OK. I'm also going to wrap the build with `/usr/bin/time` so we can see how much memory we used.

vext01:
Splat.

There was a rogue YK_DEBUG_SECTIONS=1 too, which I killed. That was removed a few PRs back.

vext01:
Out of memory again :(

vext01:
I'm doing a local build to see how much memory we use.

ltratt:
If you want to up the limit and run it through bors and inspect the output later, that's fine with me.

vext01:
I've just seen the build use all 32GiB of RAM and start swapping.

I'm seeing if it happens on upstream master, but I speculate that I have made a memory leak or I'm holding things around too long somewhere...

ltratt:
Yeah, that definitely sounds fishy...

vext01:
The maximum resident set size of any child of `x.py` as reported by gnu time is about 3.5GiB on today's upstream.

Looks like I'm going to have to track down the memory consumption :P

vext01:
The last change fixes two problems:
 * Huge memory consumption caused by duplicate node entries -- Fixed for now using sets.
 * Non-termination, which seems to have crept in at some point.

I've manually ogled some CFGs and the PHIs appear correct (but we will certainly need tests still -- next PR).

Yesterday we talked about the possibility of using bit sets instead of hash sets. Shall we do that now in this PR?

ltratt:
I've used https://crates.io/crates/fnv in grmtools -- fxhash seems unmaintained? Anyway, if all we're doing is pushing small-ish unsigned integers into a hashset, a https://crates.io/crates/bit-set will almost certainly be much faster. Is that worth looking into here?

vext01:
> fxhash seems unmaintained?

`FxHashSet` is maintained as part of the compiler.

From the fxhash landing page:
> This hashing algorithm was extracted from the Rustc compiler. 



> Anyway, if all we're doing is pushing small-ish unsigned integers into a hashset, a https://crates.io/crates/bit-set will almost certainly be much faster. Is that worth looking into here?

Happy to use a bitset, but let's use the existing bitset structures already used by the compiler?
https://github.com/rust-lang/rust/blob/master/src/librustc_data_structures/bit_set.rs

ltratt:
Yes, the compiler's bitset is what forms the basis of the bit-set crate, so that's fine with me.

vext01:
I think this should do the trick.

Just building and testing now.

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -243,40 +249,43 @@ impl<'a, 'tcx> PhiInserter<'a, 'tcx> {
...
 
         // We first need a mapping from block to the variables it defines. Appel calls this
         // `A_{orig}`. We can derive this from our definition sites.
-        let (a_orig, num_blks) = {
+        let a_orig = {
             let ykpack::Pack::Mir(ykpack::Mir{ref blocks, ..}) = self.pack;
-            let mut a_orig: IndexVec<BasicBlock, FxHashSet<TirLocal>> =
-                IndexVec::with_capacity(blocks.len());
-            a_orig.resize(blocks.len(), FxHashSet::default());
+            debug_assert!(num_tir_blks == blocks.len());
```
ltratt: This assert feels like it might be better off at line 252ish?

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -243,40 +249,43 @@ impl<'a, 'tcx> PhiInserter<'a, 'tcx> {
...
+
+            let mut a_orig: IndexVec<BasicBlock, BitSet<TirLocal>> =
```
ltratt, line 260: If you want maximum performance (but maybe this is over-optimising...), you want a single bitset of `num_tir_blks * num_tir_vars`, skipping the vec entirely.

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -454,9 +463,10 @@ impl<'tcx> ToPack<ykpack::Rvalue> for (&ConvCx<'_, 'tcx, '_>, &Rvalue<'tcx>) {
...
 }
 
-/// At the time of writing, you can't pop from a `HashSet`.
-fn hashset_pop<T>(h: &mut FxHashSet<T>) -> T where T: Eq + Hash + Clone {
-    //let e: () = h.iter().cloned().next().unwrap();
-    let e = h.iter().next().unwrap().clone();
-    h.take(&e).unwrap()
+/// At the time of writing, you can't pop from a `BitSet`.
+fn bitset_pop<T>(s: &mut BitSet<T>) -> T where T: Eq + Idx + Clone {
+    let e = s.iter().next().unwrap().clone();
```
ltratt, line 467: Does `iter` give you the next set bit, or just the next bit?

src/librustc_yk_sections/mir_cfg.rs
vext01: Ah yeah, it's possible. Shall we put a comment and come back to it if we find we need it?

src/librustc_yk_sections/mir_cfg.rs
vext01: The docstring says:
> Iterates over the indices of set bits in a sorted order.

Which I think is correct. Do you agree?

src/librustc_yk_sections/mir_cfg.rs
vext01: I think that was removed in the last commit, where I actually count the TIR blocks instead of getting the number of blocks from the MIR.

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -250,13 +250,12 @@ impl<'a, 'tcx> PhiInserter<'a, 'tcx> {
...
         let mut df = DominatorFrontiers::new(self.mir);
         let num_tir_vars = self.def_sites.len();
-        let num_tir_blks = self.mir.basic_blocks().len(); // There's one TIR block per MIR block.
 
         // We first need a mapping from block to the variables it defines. Appel calls this
         // `A_{orig}`. We can derive this from our definition sites.
-        let a_orig = {
+        let (a_orig, num_tir_blks) = {
             let ykpack::Pack::Mir(ykpack::Mir{ref blocks, ..}) = self.pack;
-            debug_assert!(num_tir_blks == blocks.len());
```
vext01: Here's where that assertion is removed.

The reason I did it this way was because, at some point we may move away from one MIR block per TIR block. In that case, this code remains correct.

[I'm hoping we don't have to make that move, as it would mean we'd have to implement our own dominators and dominator tree structures, instead of relying on the ones from the compiler, but that's another story]

src/librustc_yk_sections/mir_cfg.rs
ltratt: Agreed.

src/librustc_yk_sections/mir_cfg.rs
vext01: Actually, wait, I'm not sure I follow. The vector is acting as a mapping between basic block index and a collection of TIR variables.

I think you are suggesting we could have a single: `BitSet<(BasicBlock, TirLocal)>`?

How would you look up the locals for any given basic block without iterating the whole set?

Either way, I think we can probably leave this as-is and optimise on-demand.

ltratt: Ah, I didn't explain things very well. Oops!

What I meant was having a BitVec of length `num_tir_blks * num_tir_vars`. Since all we're doing is setting / checking individual bits, this will be *very* fast.

src/librustc_yk_sections/mir_cfg.rs
ltratt: I'm not sure I understand, but it's not that important either way.

src/librustc_yk_sections/mir_cfg.rs
vext01: Can you show me the proposed type of `a_orig` and how you'd fetch the variables for (e.g.) block index 42?

vext01:
Hey Laurie. That last commit is a **rough** attempt at abstracting so that the algorithm still looks like the one in the Appel book, while using a single bit vector for each mapping.

One thing that crosses my mind. Currently `VobMapSet` (terrible name, needs a better one) works only on `usize`s meaning the consumer has to do casting. We *could* try to make it generic over the `TryInto<usize>` (using `PhantomData` since there are not really any numbers stored inside). Then the consumer wouldn't need to care about casting.

My questions to you now:
 * Do you like where this is going?
 * Shall we make the data structure generic?

If you do like the direction, depending on the answer to the latter question, I'd still need to go back and check the casting logic and re-comment those.

Alternatively, we can cut our losses and go back to vectors of bitsets 🔪 It wouldn't hugely matter, since this isn't a performance critical section.

Cheers

vext01:
[needs final testing btw, building now]

ltratt:
I'm a bit surprised by all the complexity: don't we know exactly how big the bitvec needs to be when we create it? If so, why have we got all this resizing for? Maybe I'm missing out on something!

[I also thought you were going to use `BitVec` -- not that I mind you using `Vob`! -- but because I thought `BitVec` was already part of rustc.]

vext01:
> don't we know exactly how big the bitvec needs to be when we create it?

In one case we do not. I've made the assumption that it may not be the case that there's one TIR variable per MIR variable. I think that would be too restrictive. Hence the resizing when we are building up our definition sites mapping during conversion to a pack.

> I also thought you were going to use BitVec

I can't find a structure of this (or similar) name in the compiler or std lib.

ltratt:
It might be called `bit-vec` or `bit_vec`.



>>    don't we know exactly how big the bitvec needs to be when we create it?
>In one case we do not.

Which case?

vext01:
> Which case?

Like I said above. It's when we are building up our definition sites mapping. We don't know how many TIR variable there will be.

> It might be called bit-vec or bit_vec.

Alas not.

I found this comment in the compiler:
> // Variants is a BitVec of indexes into adt_def.variants.

But there is no further references to `BitVec` so I think it must have been removed.

ltratt:
> Like I said above. It's when we are building up our definition sites mapping.

Can you point it out to me? Maybe I'm just being dense!

Weirdly, I can see bit-vec in the `Cargo.lock`:

```
$ rg bit-vec
Cargo.lock
119: "bit-vec 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
123:name = "bit-vec"
3991:"checksum bit-vec 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)" = "4440d5cb623bb7390ae27fec0bb6c61111969860f8e3ae198bfa0663645e67cf"
```

but can't see it (or `bit_vec`) being used?!

vext01:
> Weirdly, I can see bit-vec in the Cargo.lock

That'll be a transitive dependency then?

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -27,11 +26,84 @@ use std::fs::File;
...
 
 const SECTION_NAME: &'static str = ".yk_tir";
 const TMP_EXT: &'static str = ".yk_tir.tmp";
 
+/// A conversion context holds the state needed to perform the conversion to (pre-SSA) TIR.
+struct ConvCx<'a, 'tcx, 'gcx> {
+    /// The compiler's god struct. Needed for queries etc.
+    tcx: &'a TyCtxt<'a, 'tcx, 'gcx>,
+    /// Maps TIR variables to their definition sites.
+    def_sites: RefCell<VobMapSet>,
```
vext01: This is the mapping we can't know the size of ahead of time.

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -27,11 +26,84 @@ use std::fs::File;
...
+}
+
+impl<'a, 'tcx, 'gcx> ConvCx<'a, 'tcx, 'gcx> {
+    fn new(tcx: &'a TyCtxt<'a, 'tcx, 'gcx>, mir: &Mir<'tcx>) -> Self {
+        let var_map = IndexVec::new();
+        let num_blks = mir.basic_blocks().len();
+
+        Self {
+            tcx,
+            def_sites: RefCell::new(VobMapSet::new(0, num_blks)),
```
vext01: We instantiate it empty here.

src/librustc_yk_sections/mir_cfg.rs
```diff
@@ -27,11 +26,84 @@ use std::fs::File;
...
+        self.def_sites.into_inner()
+    }
+
+    /// Add `bb` as a definition site of the TIR variable `var`.
+    fn push_def_site(&self, bb: BasicBlock, var: TirLocal) {
+        let mut sites = self.def_sites.borrow_mut();
+        // This conversion is safe because `var` was generated by `tir_var()` which guarantees that
+        // a `u32` can fit in a `usize`.
+        let var_usize = var as usize;
+        sites.insert_maybe_resize(var_usize, bb.as_usize());
```
vext01: And we insert here with a resize if necessary.

src/librustc_data_structures/vob_map_set.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// Insert `val` into the set for key `key`, resizing the map if necessary.
+    pub fn insert_maybe_resize(&mut self, key: usize, val: usize) {
+        let required_size = key.checked_add(1).unwrap().checked_mul(self.set_size).unwrap();
+        if self.vob.len() < required_size {
+            self.vob.resize(required_size, false);
+        }
+        self.insert(key, val);
+    }
+
+    /// Insert `val` into the set for key `key`. `v` must not be larger than the set size.
```
ltratt: Wrong comment?

src/librustc_data_structures/vob_map_set.rs
```diff
@@ -0,0 +1,247 @@
...
+
+/// Conceptually this is a `Map<usize, Set<usize>>`, using a single bit vector internally.
+///
+/// * The size of the sets for each key must be the same and known ahead of time.
+/// * Operations will panic if you try to access out of bounds keys or set items.
+/// * When the map is resized, new keys will be instantiated with empty sets. As such, this data
+///   structure cannot reliably express the absence of a key.
+pub struct VobMapSet {
+    vob: Vob,
+    set_size: usize,
```
ltratt: Isn't this always the same size as `vob.len()`?

src/librustc_data_structures/vob_map_set.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub fn contains(&mut self, key: usize, val: usize) -> bool {
+        self.vob.get(Self::idx(self.set_size, key, val)).unwrap()
+    }
+
+    pub fn pop(&mut self, key: usize) -> Option<usize> {
+        let start = Self::idx(self.set_size, key, 0);
+        let mut itr = self.vob.iter().skip(start);
+        for v in 0..self.set_size {
+            if itr.next().unwrap() {
+                self.vob.set(start.checked_add(v).unwrap(), false);
```
ltratt: You don't need `checked_add` here, as it clearly can't wrap (otherwise the Vob wouldn't have been created).

src/librustc_data_structures/vob_map_set.rs
```diff
@@ -0,0 +1,247 @@
...
+        self.insert(key, val);
+    }
+
+    /// Insert `val` into the set for key `key`. `v` must not be larger than the set size.
+    pub fn contains(&mut self, key: usize, val: usize) -> bool {
+        self.vob.get(Self::idx(self.set_size, key, val)).unwrap()
+    }
+
+    pub fn pop(&mut self, key: usize) -> Option<usize> {
+        let start = Self::idx(self.set_size, key, 0);
```
ltratt: Isn't this roughly equivalent to something like:

```rust
self.vob.iter_set_bits(self.set_size * ..., self.set_size * ...).next().unwrap_or(None)
```

?

ltratt:
Having looked at this, I'm not sure this is an improvement overall :/ It will be faster, but we've had to add a lot of code for that, and I'm not sure it's worth the maintenance headache. This is clearly my fault for suggesting this in the first place, so sorry for that!

Maybe the `Vec<BitSet>` approach was the right compromise between speed and simplicity?

src/librustc_data_structures/vob_map_set.rs
vext01: No. It's the size of the set for each key. `vob.len()` will be this number multiplied by the number of keys.

src/librustc_data_structures/vob_map_set.rs
vext01: Oops. Yeah!

src/librustc_data_structures/vob_map_set.rs
ltratt: OK.

src/librustc_data_structures/vob_map_set.rs
vext01: Ah ha! Yes, that'd be neater.

src/librustc_data_structures/vob_map_set.rs
vext01: OK, but it looks like this code will disappear due to your previous suggestion.

vext01:
> This is clearly my fault for suggesting this in the first place, so sorry for that!

No worries. No harm done. At least we can say we tried :)

> Maybe the Vec<BitSet> approach was the right compromise between speed and simplicity?

If this code was in the performance critical path, I think we might have been doing the right thing, but in this case, I can  revert the last change without losing sleep.

Sounds like a plan?

ltratt:
Yes, I think reverting is the right thing. Sorry :(

vext01:
Is it squash time now?

ltratt:
Please squash.

vext01:
Squashed and revised commit message.

PR #12, opened 2019-04-08
simpler dominance frontiers

vext01:
These changes are optimisations and simplifications to the machinery around computing dominance frontiers and inserting phi nodes.

Most notably instead of using the frontier algorithm from Appel book, we instead use the  one from the paper "A Simple,Fast Dominance Algorithm. Keith D. Cooper, Timothy J. Harvey, and Ken Kennedy":
https://www.doc.ic.ac.uk/~livshits/classes/CO444H/reading/dom14.pdf

The Rust compiler already implements most of this paper, I just had to add the last little bit.

I had been using grmtools as a benchmark for these changes and was underwhelmed when I saw that they made no performance difference whatsoever :( I was going to suggest we include these changes anyway, as they make the code shorter and easier to read.

But then I tried with these optimisations on the compiler itself and found it shaved off about 2 hours and 15 minutes! I thought it was a fluke, so I repeated the build another two times over the weekend. Sure enough, consistent speedup!

http://bencher12.soft-dev.org:8010/#/builders/2/builds/117
http://bencher12.soft-dev.org:8010/#/builders/2/builds/118
http://bencher12.soft-dev.org:8010/#/builders/2/builds/119

And here's a build from before where the build was slower:
http://bencher12.soft-dev.org:8010/#/builders/2/builds/110

So it was about 6 hours 15 mins, now it's about 4 hours. The build is now faster than when we had a hand-rolled serialiser.

There is one more potential optimisation: remove `RefCell`s. I did a little benchmark and found that refcells have a fair overhead if you use them in a tight loop (especially in release mode). However, to remove the `RefCell`s from our conversion code, we'd need quite a bit of re-factoring. Since I'm getting itchy to move on, I suggest we don't worry about that for now.

### DIFF
--- a/src/librustc_data_structures/graph/dominators/mod.rs
+++ b/src/librustc_data_structures/graph/dominators/mod.rs
@@ -7,6 +7,7 @@
 use super::super::indexed_vec::{Idx, IndexVec};
 use super::iterate::reverse_post_order;
 use super::ControlFlowGraph;
+use crate::bit_set::BitSet;
 
 use std::fmt;
 
@@ -156,6 +157,36 @@ impl<'dom, Node: Idx> Iterator for Iter<'dom, Node> {
         } else {
             return None;
         }
+    }
+}
+
+pub struct DominatorFrontiers<G: ControlFlowGraph> {
+    dfs: IndexVec<G::Node, BitSet<G::Node>>,
+}
+
+impl<G: ControlFlowGraph> DominatorFrontiers<G> {
+    pub fn new(graph: &G, doms: &Dominators<G::Node>) -> Self {
+        let num_nodes = graph.num_nodes();
+        let mut dfs = IndexVec::from_elem_n(BitSet::new_empty(num_nodes), num_nodes);
+
+        for b in (0..num_nodes).map(|i| G::Node::new(i)) {
+            if graph.predecessors(b).take(2).count() > 1 {
+                // Iterator isn't clonable, so we have to make a new one.
+                let preds = graph.predecessors(b);
+                for p in preds {
+                    let mut runner = p; // Not strictly necessary, but for parity with the paper.
+                    while runner != doms.immediate_dominator(b) {
+                        dfs[runner].insert(b);
+                        runner = doms.immediate_dominator(runner);
+                    }
+                }
+            }
+        }
+       Self { dfs }
+    }
+
+    pub fn frontier(&self, n: G::Node) -> &BitSet<G::Node> {
+        &self.dfs[n]
     }
 }
 
--- a/src/librustc_yk_sections/emit_tir.rs
+++ b/src/librustc_yk_sections/emit_tir.rs
@@ -31,7 +31,7 @@ use std::cell::{Cell, RefCell};
 use std::mem::size_of;
 use rustc_data_structures::bit_set::BitSet;
 use rustc_data_structures::indexed_vec::{IndexVec, Idx};
-use rustc_data_structures::graph::dominators::Dominators;
+use rustc_data_structures::graph::dominators::DominatorFrontiers;
 use ykpack;
 use ykpack::LocalIndex as TirLocal;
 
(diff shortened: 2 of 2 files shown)


src/librustc_data_structures/graph/dominators/mod.rs
```diff
@@ -159,6 +160,36 @@ impl<'dom, Node: Idx> Iterator for Iter<'dom, Node> {
...
+    dfs: IndexVec<G::Node, BitSet<G::Node>>,
+}
+
+impl<G: ControlFlowGraph> DominatorFrontiers<G> {
+    pub fn new(graph: &G, doms: &Dominators<G::Node>) -> Self {
+        let num_nodes = graph.num_nodes();
+        let mut dfs = IndexVec::from_elem_n(BitSet::new_empty(num_nodes), num_nodes);
+
+        for b in (0..num_nodes).map(|i| G::Node::new(i)) {
+            if graph.predecessors(b).take(2).count() > 1 {
```
ltratt, line 173: What's the maximum number of predecessors? Or is there no upper bound?

vext01: Here we only care if it's a join point -- i.e. it has more than one predecessor.

In general there can be unbounded predecessors.

As you can see from a later comment, this code isn't as nice as I'd hoped, but hey.

src/librustc_yk_sections/emit_tir.rs
```diff
@@ -208,88 +208,66 @@ fn do_generate_tir<'a, 'tcx, 'gcx>(
...
+               mut def_sites: Vec<BitSet<BasicBlock>>)
+{
+    let doms = mir.dominators();
+    let df = DominatorFrontiers::new(mir, &doms);
+    let num_tir_vars = def_sites.len();
+
+    // We first need a mapping from block to the variables it defines. Appel calls this
+    // `A_{orig}`. We can derive this from our definition sites.
+    let (a_orig, num_tir_blks) = {
+        let num_tir_blks = blocks.len();
```
ltratt: This variable name is as long as the RHS expression. I'd personally remove the variable and just use the RHS expression where needed.

src/librustc_data_structures/graph/dominators/mod.rs
vext01: image

src/librustc_yk_sections/emit_tir.rs
vext01: That code is removed in a later commit.

src/librustc_data_structures/graph/dominators/mod.rs
ltratt: I was partly wondering if there's a nicer way of saying "how many predecessors does `b` have"? rather than firing up an iterator. It might not be worth worrying about (depending on how expensive the iterator is to set up).

src/librustc_yk_sections/emit_tir.rs
ltratt: OK.

src/librustc_data_structures/graph/dominators/mod.rs
vext01: Yeah, I was hoping for this too!

Predecessors are expressed by a trait with a sole method which returns an iterator:
https://github.com/rust-lang/rust/blob/3750348daff89741e3153e0e120aa70a45ff5b68/src/librustc_data_structures/graph/mod.rs#L35-L43

And then the MIR implementation of this trait is in terms of `predecessors_for()`:
https://github.com/rust-lang/rust/blob/3750348daff89741e3153e0e120aa70a45ff5b68/src/librustc/mir/mod.rs#L233-L236

Let me have a look if I can use any of this stuff directly, but I might be hard to access as we are behind parameterised types.

vext01: [Another option is to `collect()` the predecessors]

vext01: Yeah, so at the point where we are fiddling with those iterators, we are working with a `ControlFlowGraph`, which a `MIR` is an instance of. We can't get directly at the underlying `predecessors_for()` of the `MIR` struct, I'm afraid.

ltratt: OK, then I think this is as good as it gets. Using `collect` would almost certainly be worse than what we have.

vext01: We *could* implement this function (elsewhere, probably in `emit_tir.rs`) as a concrete function over a MIR of course. I only implemented it here because the rest of the code from that paper was also here and already generic.

ltratt: It's not worth it. I just wanted to check if something easy had been missed.

vext01: OK.

PR #16, opened 2019-04-25
Really install cargo

vext01:

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -4,21 +4,26 @@
 
 set -e
 
+export RUST_BACKTRACE=1
+export PATH=PATH=/opt/gdb-8.2/bin:${PATH}
+
+TARBALL_TOPDIR=`pwd`/build/ykrustc-stage2-latest
+TARBALL_NAME=ykrustc-stage2-latest.tar.bz2
+SNAP_DIR=/opt/ykrustc-bin-snapshots
+
 # Ensure the build fails if it uses excessive amounts of memory.
 ulimit -d $((1024 * 1024 * 8)) # 8 GiB
 
 # Note that the gdb must be Python enabled.
-/usr/bin/time -v env PATH=/opt/gdb-8.2/bin:${PATH} \
-    RUST_BACKTRACE=1 ./x.py test --config .buildbot.toml
+/usr/bin/time -v ./x.py test --config .buildbot.toml
+
+# Build extended tools and install into TARBALL_TOPDIR.
+mkdir -p ${TARBALL_TOPDIR}
+/usr/bin/time -v ./x.py install --config .buildbot.toml
 
 # Archive the build and put it in /opt
-TARBALL_TOPDIR=ykrustc-stage2
-TARBALL_NAME=ykrustc-stage2-latest.tar.bz2
-SNAP_DIR=/opt/ykrustc-bin-snapshots
-
-cd build/x86_64-unknown-linux-gnu
-ln -sf stage2 ${TARBALL_TOPDIR}
 git show -s HEAD > ${TARBALL_TOPDIR}/VERSION
-tar hjcvf ${TARBALL_NAME} ${TARBALL_TOPDIR}
+cd build
+tar jcf ${TARBALL_NAME} `basename ${TARBALL_TOPDIR}`
 chmod 775 ${TARBALL_NAME}
 mv ${TARBALL_NAME} ${SNAP_DIR} # Overwrites any old archive.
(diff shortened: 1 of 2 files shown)


vext01:
bors try

.buildbot.sh
```diff
@@ -12,13 +12,13 @@ ulimit -d $((1024 * 1024 * 8)) # 8 GiB
     RUST_BACKTRACE=1 ./x.py test --config .buildbot.toml
 
 # Archive the build and put it in /opt
-TARBALL_TOPDIR=ykrustc-stage2
+TARBALL_TOPDIR=ykrustc-stage2 # To match [install] in the build config.
```
ltratt: This comment is a bit too cryptic, I think.

.buildbot.sh
```diff
@@ -12,13 +12,13 @@ ulimit -d $((1024 * 1024 * 8)) # 8 GiB
...
 TARBALL_NAME=ykrustc-stage2-latest.tar.bz2
 SNAP_DIR=/opt/ykrustc-bin-snapshots
 
-cd build/x86_64-unknown-linux-gnu
-ln -sf stage2 ${TARBALL_TOPDIR}
+mkdir -p ${TARBALL_TOPDIR} # Or you get "cannot canonicalize..."
+env RUST_BACKTRACE=1 ./x.py install --config .buildbot.toml
```
ltratt: Do we need the `RUST_BACKTRACE` thing on? IIRC, it does have some sort of performance consequence.

vext01: I found a post where someone claims it does, but no-one knows why.

I'd like to have backtraces if possible, because otherwise I may have to build again to know what went wrong.

.buildbot.sh
vext01: It has to match the string used in the toml file:
```
[install]
prefix = ...
```

vext01: (i'll push a better comment once it works)

vext01:
This worked, but we added about an hour to the build time.

The last pushes (amongst a couple of other bits) prevents rust from building documentation. Hopefully it'll be faster.

bors try

vext01:
```
+ git show -s HEAD^2
fatal: ambiguous argument 'HEAD^2': unknown revision or path not in the working tree.
```

That is caused by one of the other commits.

So it looks like we might shave 15 minutes off without building docs.

vext01:
Looking at the logs, the `install` stage builds:
 * the installer (whatever that is)
 * cargo
 * rustfmt

And those three items take:
```
Build completed successfully in 0:49:36
```

So I think that's all time accounted for.

We might be able to invoke individual targets for `cargo` and `rustfmt` and then *not* use `cargo install` to avoid building the installer.

Do you have any thoughts before I go ahead with that?

ltratt:
Let's do it.

vext01:
bors try

vext01:
bors try

vext01:
I don't understand why that took *longer* than what we had before.

ltratt:
Do you want to squash?

vext01:
> Do you want to squash?

Well no, because the build took ages compared to what we had before.

I'm starting to wonder if there's a lot of variation...

I'm going to kick of the build again on the same version and see if it takes the same time.

bors try

ltratt:
OK.

vext01:
So we ran two builds on the same commit, the first took 5.5 hours, the second took 4.5. Looks to me like there's significant non-determinism in the compiler, or perhaps the buildbot was busier during the first build?

What do you think we should do? Good news is, once SSA is gone the build should be a bit faster.

ltratt:
I wouldn't read too much into the times: who knows what else the machine was doing at that point? Can we merge this now or ...?

vext01:
bors try

vext01:
> Can we merge this now or ...?

Not just yet.

I did a test run on yk, and it bombed out because rustdoc wasn't present. The last commit invokes that target. Bors is testing that now.

Another annoying quirk is that when you don't use `cargo install` (but instead invoke targets manually as we have here), `cargo fmt` doesn't work. You have to invoke `rustfmt` manually with a file list. A bit annoying...

vext01:
I'm going to revert to using `cargo install` to build the tarball because I don't want to layer on any more hacks to make this work.

ltratt:
OK.

vext01:
bors try

vext01:
OK, so this works. Let me now test the resulting tarball on yk.

vext01:
https://github.com/softdevteam/yk/pull/1 is now passing.

I think this is ready to squash, unless you have any comments?

ltratt:
Please squash.

vext01:
splat

PR #19, opened 2019-05-07
Rename file to something a little less generic.

ltratt:
`.buildbot.toml` was likely to be confused by new users as being a config for BuildBot, when it's actually a config for rustc when building on BuildBot.


ltratt:
That was silly of me. That new commit will need squashing.

vext01:
LGTM. Squash away.

ltratt:
Squashed.

PR #20, opened 2019-05-09
Implement a load more TIR lowerings, including constant integers.

vext01:
This is the companion PR for https://github.com/softdevteam/yk/pull/5

### DIFF
--- a/src/librustc/mir/interpret/value.rs
+++ b/src/librustc/mir/interpret/value.rs
@@ -332,6 +332,13 @@ impl<'tcx, Tag> Scalar<Tag> {
         Ok(b as u8)
     }
 
+    pub fn to_u16(self) -> EvalResult<'static, u16> {
+        let sz = Size::from_bits(16);
+        let b = self.to_bits(sz)?;
+        assert_eq!(b as u16 as u128, b);
+        Ok(b as u16)
+    }
+
     pub fn to_u32(self) -> EvalResult<'static, u32> {
         let sz = Size::from_bits(32);
         let b = self.to_bits(sz)?;
@@ -344,6 +351,11 @@ impl<'tcx, Tag> Scalar<Tag> {
         let b = self.to_bits(sz)?;
         assert_eq!(b as u64 as u128, b);
         Ok(b as u64)
+    }
+
+    pub fn to_u128(self) -> EvalResult<'static, u128> {
+        let sz = Size::from_bits(128);
+        Ok(self.to_bits(sz)?)
     }
 
     pub fn to_usize(self, cx: &impl HasDataLayout) -> EvalResult<'static, u64> {
@@ -360,6 +372,14 @@ impl<'tcx, Tag> Scalar<Tag> {
         Ok(b as i8)
     }
 
+    pub fn to_i16(self) -> EvalResult<'static, i16> {
+        let sz = Size::from_bits(16);
+        let b = self.to_bits(sz)?;
+        let b = sign_extend(b, sz) as i128;
+        assert_eq!(b as i16 as i128, b);
+        Ok(b as i16)
+    }
+
     pub fn to_i32(self) -> EvalResult<'static, i32> {
         let sz = Size::from_bits(32);
         let b = self.to_bits(sz)?;
@@ -374,6 +394,12 @@ impl<'tcx, Tag> Scalar<Tag> {
         let b = sign_extend(b, sz) as i128;
         assert_eq!(b as i64 as i128, b);
         Ok(b as i64)
+    }
+
+    pub fn to_i128(self) -> EvalResult<'static, i128> {
+        let sz = Size::from_bits(128);
+        let b = self.to_bits(sz)?;
+        Ok(sign_extend(b, sz) as i128)
     }
 
     pub fn to_isize(self, cx: &impl HasDataLayout) -> EvalResult<'static, i64> {
--- a/src/librustc_yk_sections/Cargo.toml
+++ b/src/librustc_yk_sections/Cargo.toml
@@ -16,3 +16,4 @@ rustc_yk_link = { path = "../librustc_yk_link" }
 ykpack = { git = "https://github.com/softdevteam/yk" }
 rustc_codegen_utils = { path = "../librustc_codegen_utils" }
 rustc_data_structures = { path = "../librustc_data_structures" }
+syntax = { path = "../libsyntax" }
--- a/src/librustc_yk_sections/emit_tir.rs
+++ b/src/librustc_yk_sections/emit_tir.rs
@@ -17,10 +17,11 @@ use rustc::hir::def_id::DefId;
 use rustc::mir::{
     Mir, TerminatorKind, Operand, Constant, StatementKind, BasicBlock, BasicBlockData, Terminator,
     Place, Rvalue, Statement, Local, PlaceBase, BorrowKind, BinOp, UnOp, NullOp, Projection,
-    AggregateKind
+    AggregateKind, ProjectionElem
 };
-use rustc::ty::{TyS, TyKind, Const, LazyConst};
+use rustc::ty::{TyS, TyKind, Const, LazyConst, Ty};
 use rustc::util::nodemap::DefIdSet;
+use syntax::ast::{IntTy, UintTy};
 use std::path::PathBuf;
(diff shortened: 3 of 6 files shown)


src/librustc/mir/interpret/value.rs
```diff
@@ -346,6 +353,12 @@ impl<'tcx, Tag> Scalar<Tag> {
         Ok(b as u64)
     }
 
+    pub fn to_u128(self) -> EvalResult<'static, u128> {
+        let sz = Size::from_bits(128);
+        let b = self.to_bits(sz)?;
+        Ok(b)
```
ltratt: The intermediate variable `b` seems pointless.

src/librustc/mir/interpret/value.rs
```diff
@@ -332,6 +332,13 @@ impl<'tcx, Tag> Scalar<Tag> {
         Ok(b as u8)
     }
 
+    pub fn to_u16(self) -> EvalResult<'static, u16> {
+        let sz = Size::from_bits(16);
+        let b = self.to_bits(sz)?;
+        assert_eq!(b as u16 as u128, b);
```
ltratt, line 338: Why is this assertion necessary? I also think that `from_bits` or `to_bits` should probably be doing the cast to the right type? Or am I missing out on something?

src/librustc/mir/interpret/value.rs
vext01: Ah, it is. It's because I was copy/paste/editting similar code from rustc.

src/librustc/mir/interpret/value.rs
vext01: It's copied from the other accessors already in Rustc.

src/librustc_yk_sections/Cargo.toml
```diff
@@ -13,6 +13,8 @@ test = false
 [dependencies]
 rustc = { path = "../librustc" }
 rustc_yk_link = { path = "../librustc_yk_link" }
-ykpack = { git = "https://github.com/softdevteam/yk" }
+#ykpack = { git = "https://github.com/softdevteam/yk" }
+ykpack = { path = "/home/vext01/research/yk/ykpack" }
```
ltratt: I know why this is here, but we need to work out something better than leaking private pathnames out each time. Unfortunately I'm not sure what that better thing might be. We could point `ykpack` at a `dev` branch or something?

src/librustc/mir/interpret/value.rs
vext01: (They had existing implementations for all numeric types apart from u16, u128, i16, i128)

src/librustc_yk_sections/emit_tir.rs
```diff
@@ -508,3 +538,79 @@ impl<'tcx> ToPack<ykpack::UnOp> for (&ConvCx<'_, 'tcx, '_>, &UnOp) {
...
+    }
+}
+
+//fn bits_to_pack<T>(k
+
+/// Ty + LazyConst -> Pack
+impl<'tcx> ToPack<ykpack::Constant> for (&ConvCx<'_, 'tcx, '_>, Ty<'tcx>, &LazyConst<'tcx>) {
+    fn to_pack(&mut self) -> ykpack::Constant {
+        let (ccx, ty, lconst) = self;
+        if let LazyConst::Evaluated(cst) = lconst {
```
ltratt: This might be simpler as a `match`?

src/librustc_yk_sections/Cargo.toml
vext01: It doesn't bother me too much. There's nothing secret in the path name. I'd rather not complicate the git situation more than necessary.

src/librustc_yk_sections/emit_tir.rs
```diff
@@ -508,3 +538,79 @@ impl<'tcx> ToPack<ykpack::UnOp> for (&ConvCx<'_, 'tcx, '_>, &UnOp) {
...
+            (*ccx, *ty, cst).to_pack()
+        } else {
+            ykpack::Constant::Unimplemented
+        }
+    }
+}
+
+/// Ty + Const -> Pack
+impl<'tcx> ToPack<ykpack::Constant> for (&ConvCx<'_, 'tcx, '_>, Ty<'tcx>, &Const<'tcx>) {
```
ltratt, line 568: Dumb question: why are we throwing around triples like this rather than having a named struct?

src/librustc_yk_sections/emit_tir.rs
vext01: Agreed.

src/librustc/mir/interpret/value.rs
ltratt: It feels a bit icky, and on some platforms (which have slow `u128` support) also quite slow. Maybe it's not worth worrying about though.

src/librustc_yk_sections/emit_tir.rs
vext01: OK, this is hard to explain.

In the beginning, we were passing around only pairs of `(ccx, X)`, where `X` is the thing we want to lower and `ccx` carries state, like the next free TIR variable number and the compiler's `tcx` (needed for various queries).

As we diverged from MIR further, the need arose to refer to things we saw earlier in the lowering chain. In these cases those pairs became triples.

For example, in the above, we need to know the type of the MIR `Const` we are converting so that we can do an appropriate lowering to one of our TIR constant enums. The type arose a few stages earlier in the chain, so we have to pass it down.

We could define structs, but I recommend we don't. To be consistent, we'd need one for every kind of thing we want to lower. We currently have 22 lowerings and more will come. I also think it will be harder to understand, so let's leave it be.

ltratt: I must admit, something here doesn't feel right to me. If you think about an RPython trace, it's very simple in terms of the variety of things that can be in it, and we're going to want something similar. I suspect we're passing too much of MIR's complexity through to the run-time. Warning: I might be very wrong.

src/librustc/mir/interpret/value.rs
vext01: I don't know. I was just following the precedent:
https://github.com/rust-lang/rust/blob/9f8396158457943b63bb39534b9ef86868f7cfac/src/librustc/mir/interpret/value.rs#L339-L395

ltratt: OK, let's leave it as-is.

src/librustc_yk_sections/emit_tir.rs
vext01: I don't think you are wrong, but I'm also not sure what that has to do with passing the triples down at compile time.

I think we can certainly simplify our TIR, but in stages. I was going to try and simplify assignments in the next PR.

ltratt: OK, fair enough. Let's make the next PR purely about simplification then?

vext01: Sure.

ltratt:
OK, so we've mostly resolved things, with one or two minor things to change (e.g. `if let` to `match`).

vext01:
Comments addressed. First we need to merge:
https://github.com/softdevteam/yk/pull/5

ltratt:
You can squash this (fixing the path at the same time) while we wait for that.

vext01:
Splat.

vext01:
Toying with disabling the external crates whitelist, or at-least always allowing ykpack... thoughts?

ltratt:
Please squash.

vext01:
splat

vext01:
Can we retry this? It looks like buildbot failed to pick up the job and bors timed out.

ltratt:
That's the first time that's happened. We might need to keep an eye on that!

bors r+

vext01:
This is weird. Buildbot picked up the job and it failed (a try_from failed), but github didn't see it.

I'm going to restart buildbot, but in light of what we now now about how TIR needs to be, I wonder if it's even worth fixing this...

ltratt:
It's probably worth quickly checking this, as it doesn't sound like an issue specific to this PR? But if it's difficult, we can defer it until we encounter it again.

vext01:
Closing

PR #21, opened 2019-05-13
Remove most of the first attempt of TIR lowering.

vext01:
This is a companion to https://github.com/softdevteam/yk/pull/6.

I've left some unused code behind that I know I will need to adapt later, hence the `#![allow(...)]`.

One the ykpack change is merged, I'll need to bump the version in the whitelist.


src/librustc_yk_sections/emit_tir.rs
```diff
@@ -11,15 +11,14 @@
 //!
 //! Serialisation itself is performed by an external library: ykpack.
 
+// FIXME, just temporarily.
```
ltratt, line 14: Why?

vext01: > I've left some unused code behind that I know I will need to adapt later, hence the #![allow(...)].

ltratt: Yes, but what code? And can we briefly explain why it's temporary? I always worry these things become more permanent than expected!

vext01: Every field in the conversion context is currently unused, but we will certainly use them later.

I went the `allow` route instead of underscoring all the variable names, as I thought it would be easier to review.

vext01: To expand on that:

For example the next variable counter is unused since there are no conversions to make new variables.

ltratt: Is this going to be fixed very soon (e.g .the next PR)?

vext01: I hope so, yes!

vext01:
I've added a couple of commits.

The first makes the tidy check allow our repos without an explicit version in the whitelist.

The second updates the Cargo.lock. Sadly it looks like this is going to be a step that we have to manually do each time we merge a ykpack change.

ltratt:
I don't understand what this is doing: what does whitelisting the tidy check actually do?

vext01:
> I don't understand what this is doing: what does whitelisting the tidy check actually do?

Part of the tidy checker's job is (for whatever upstream's reason) to check that only "allowed" sources are used in Rustc.

All of crates.io is allowed. Anything else needs an explicit entry in a whitelist.

Because whitelist entries are versioned, each time one of our dependencies (e.g. ykpack) is updated, we have to update the version in the whitelist. Failing to do so makes `cargo test` fail. I find this annoying.

My change unconditionally whitelists our team github account.

ltratt:
Aha, OK, I understand now. I'm deeply offended that our account wasn't whitelisted by rustc earlier ;)

Time to squash?

vext01:
Yep, squashed.

PR #22, opened 2019-05-21
Start lowering to the new TIR format.

vext01:
Here's the beginnings of lowering to the new TIR format from https://github.com/softdevteam/yk/pull/7

Only some assignments are lowered currently.

Note that I ditched the ToPack trait we used before in favour of just using self functions. This means fewer explicit lifetimes have to be specified.

Laurie can sleep easy now this stuff is gone: :P
```
#![allow(unused_variables)]
#![allow(dead_code)]
```

### DIFF
--- a/src/librustc_yk_sections/emit_tir.rs
+++ b/src/librustc_yk_sections/emit_tir.rs
@@ -11,14 +11,12 @@
 //!
 //! Serialisation itself is performed by an external library: ykpack.
 
-// FIXME, just temporarily.
-#![allow(unused_variables)]
-#![allow(dead_code)]
-
 use rustc::ty::TyCtxt;
 
 use rustc::hir::def_id::DefId;
-use rustc::mir::{Mir, Local};
+use rustc::mir::{
+    Mir, Local, BasicBlockData, Statement, StatementKind, Place, PlaceBase, Rvalue, Operand
+};
 use rustc::util::nodemap::DefIdSet;
 use std::path::PathBuf;
 use std::fs::File;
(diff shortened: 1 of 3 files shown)


ltratt:
Looks OK to me.

vext01:
This is ready for bors if Jake is happy.

src/librustc_yk_sections/emit_tir.rs
```diff
@@ -49,35 +46,37 @@ struct ConvCx<'a, 'tcx, 'gcx> {
     /// The compiler's god struct. Needed for queries etc.
     tcx: &'a TyCtxt<'a, 'tcx, 'gcx>,
     /// Monotonically increasing number used to give TIR variables a unique ID.
-    next_tir_var: Cell<TirLocal>,
+    next_tir_var: Cell<ykpack::LocalIndex>,
     /// A mapping from MIR variables to TIR variables.
-    var_map: RefCell<IndexVec<Local, Option<TirLocal>>>,
```
jacob-hughes, line 54: I'm curious why this is a `RefCell`. From what I can see, it's only mutated as part of `tir_val()`. Did you have difficulties fighting the borrow checker accepting a `&mut self` reference to `ConvCx` here? Or was it conscious design decision?

If the latter, I think I would prefer passing an explicit mutable reference to `ConvCx` over interior mutability. It's easier to tell that calling this function _will_ cause an internal state change.

jacob-hughes: I commented this on the minus diff by mistake, but I meant it on the new `var_map` field def too ;)

vext01: This was a good spot. The `RefCell` is left over from before. I don't think it's needed.

Commit coming...

vext01:
The last commit removes the RefCell and a Cell too.

The lock file bumped because of the yk change being merged.

I realised I forgot to check the tests. Building now, will take some time.

vext01:
OK, building didn't take that long.

The last commit fixes the test. The test isn't very meaningful yet, but will make more sense over time.

OK to squash?

jacob-hughes:
Yes LGTM.

vext01:
splat

ltratt:
And me.

PR #37, opened 2019-08-06
Remove custom DefId collection in favour of tcx.collect_and_partition_mono_items().

vext01:
This draft implements @bjorn3's suggestion:
> you can use tcx.collect_and_partition_mono_items to get all functions and statics which need codegen instead of doing it manually. Your way may be a bit faster though.

In my rough benchmarking of a small program, we didn't slow down any measurable amount.  I'm going to `bors try` this first to see if we slowed down the compilation process too much.

bors try


ltratt:
Assuming you're happy with the performance, I'm happy with this PR.

vext01:
> Assuming you're happy with the performance, I'm happy with this PR.

I can't really conclusively say that performance was impacted at all, so I'm happy.

PR #38, opened 2019-08-16
Discover additional DefIds by looking in the def_path_table.

vext01:
The "definition path table" is a mapping from DefIds to definition
paths. There's an entry for every item than appears in the compiler's
metadata, thus making it an excellent discovery tool for our serialiser
code to find DefIds in dependent crates (if you run with `-Z
always-encode-mir`)!

We also switch to using tcx.mir_keys() for finding the DefIds in the
current crate. Previously we used tcx.collect_and_partition_mono_items()
which may miss some stuff.

A future change will remove the work list used in the serialised, as this
code supersedes it. That will be a fairly large change, so saving it for
another PR.

Another future change will enable `-Z always-encode-mir` by default for
ykrustc.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -11,7 +11,7 @@ TARBALL_NAME=ykrustc-stage2-latest.tar.bz2
 SNAP_DIR=/opt/ykrustc-bin-snapshots
 
 # Ensure the build fails if it uses excessive amounts of memory.
-ulimit -d $((1024 * 1024 * 8)) # 8 GiB
+ulimit -d $((1024 * 1024 * 12)) # 12 GiB
 
 # Note that the gdb must be Python enabled.
 /usr/bin/time -v ./x.py test --config .buildbot.config.toml
--- a/src/librustc/middle/cstore.rs
+++ b/src/librustc/middle/cstore.rs
@@ -215,6 +215,7 @@ pub trait CrateStore {
                                  tcx: TyCtxt<'a, 'tcx, 'tcx>)
                                  -> EncodedMetadata;
     fn metadata_encoding_version(&self) -> &[u8];
+    fn def_path_table_lens(&self) -> Vec<usize>;
 }
 
 pub type CrateStoreDyn = dyn CrateStore + sync::Sync;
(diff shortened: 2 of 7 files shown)


vext01:
The test failures on this one are weird. I can only assume that the use of `tcx.mir_keys` was incorrect and has messed up the MIR. Trying to revert that part back.

Will take some time to test.

vext01:
The last commit goes back to using mono partitioning and skips the two problematic tests:
 * The empty string test is (we think) just unlucky. debuginfo is only ever a best-effort kind of affair.
 * The generator test is being discussed upstream: https://github.com/rust-lang/rust/pull/63742

ltratt:
Please squash (if you wish), though please reference the upstream PR that's discussing what should happen in such cases.

vext01:
done, and splat.

vext01:
Let's r+ this?

vext01:
> memory allocation of 240 bytes failed

So we ran out of `ulimit -d`. I have an idea to fix it.

vext01:
This is troubling. The last push makes it so we only discover DefIds that actually have MIR, but this still isn't enough; A liballoc doctest is still consuming too much memory.

The memory limit we've set is 8GiB. Although it's arbitrary, I don't want to up it, as that's quite a lot of RAM already.

Before we go back to the drawing board on this, I want to check this problem definitely doesn't happen on master with bencher13 in it's current state.

vext01:
> I want to check this problem definitely doesn't happen on master with bencher13

It doesn't. So I have to think again I'm afraid.

ltratt:
rustc does take quite a lot of RAM to compile, so I'd be inclined to up it to (say) 12GiB and if it still runs out of memory then I agree: we have a problem.

vext01:
It looks like we got lucky.

Shall I squash?

ltratt:
Please squash.

vext01:
Splat

vext01:
Sorry, I pushed to the wrong repo. Now we are correctly squashed :)

vext01:
Sigh, this PR..... 💃 

I've fixed the tidy check, ok to squash again?

ltratt:
Yes please.

vext01:
Splat

PR #41, opened 2019-08-29
Insert yorick labels at the start of each block and insert compile-time tracer switch.

ptersilie:
This PR adds yorick labels to basic MIR blocks which should allow us match them against traces in the future.

Though this PR is likely not ready to be merged yet, I have opened it to make it easier to discuss what needs to be done next.

### DIFF
--- a/src/librustc_codegen_llvm/builder.rs
+++ b/src/librustc_codegen_llvm/builder.rs
@@ -5,6 +5,7 @@ use crate::context::CodegenCx;
 use crate::type_::Type;
 use crate::type_of::LayoutLlvmExt;
 use crate::value::Value;
+use crate::llvm::debuginfo::DIScope;
 use syntax::symbol::LocalInternedString;
 use rustc_codegen_ssa::common::{IntPredicate, TypeKind, RealPredicate};
 use rustc_codegen_ssa::MemFlags;
@@ -20,7 +21,7 @@ use rustc_codegen_ssa::mir::operand::{OperandValue, OperandRef};
 use rustc_codegen_ssa::mir::place::PlaceRef;
 use rustc_target::spec::{HasTargetSpec, Target};
 use std::borrow::Cow;
-use std::ffi::CStr;
+use std::ffi::{CStr, CString};
 use std::ops::{Deref, Range};
 use std::ptr;
 use std::iter::TrustedLen;
@@ -57,6 +58,7 @@ impl BackendTypes for Builder<'_, 'll, 'tcx> {
     type Funclet = <CodegenCx<'ll, 'tcx> as BackendTypes>::Funclet;
 
     type DIScope = <CodegenCx<'ll, 'tcx> as BackendTypes>::DIScope;
+    type DISubprogram = <CodegenCx<'ll, 'tcx> as BackendTypes>::DISubprogram;
 }
 
 impl ty::layout::HasDataLayout for Builder<'_, '_, '_> {
@@ -154,9 +156,48 @@ impl BuilderMethods<'a, 'tcx> for Builder<'a, 'll, 'tcx> {
         }
     }
 
+    fn position_before(&mut self, instr: &Value) {
+        unsafe {
+            llvm::LLVMPositionBuilderBefore(self.llbuilder, instr);
+        }
+    }
+
     fn position_at_end(&mut self, llbb: &'ll BasicBlock) {
         unsafe {
             llvm::LLVMPositionBuilderAtEnd(self.llbuilder, llbb);
+        }
+    }
+
+    fn add_yk_block_label(&mut self, di_sp: &DIScope, lbl_name: CString) {
+        // Insert a DWARF label at the start of each block.
+        // ykrt uses this at runtime to map virtual addresses to MIR blocks.
+        if let Some(first_instr) = self.first_instruction(self.llbb()) {
+            self.position_before(first_instr);
+
+            let dbg_cx = self.cx().dbg_cx.as_ref().unwrap();
+            let di_bldr = dbg_cx.get_builder();
+
+            unsafe {
+                llvm::LLVMRustAddYkBlockLabel(self.llbuilder, di_bldr, di_sp, first_instr,
+                                              lbl_name.as_ptr());
+            }
+        }
+
+        self.position_at_end(self.llbb());
+    }
+
+    fn add_yk_block_label_at_end(&mut self, di_sp: &DIScope, lbl_name: CString) {
+        let dbg_cx = self.cx().dbg_cx.as_ref().unwrap();
+        let di_bldr = dbg_cx.get_builder();
+        unsafe {
+            llvm::LLVMRustAddYkBlockLabelAtEnd(self.llbuilder, di_bldr, di_sp, self.llbb(),
+                                               lbl_name.as_ptr());
+        }
+    }
+
+    fn first_instruction(&mut self, llbb: &'ll BasicBlock) -> Option<&'ll Value> {
+        unsafe {
+            llvm::LLVMGetFirstInstruction(llbb)
         }
     }
 
--- a/src/librustc_codegen_ssa/mir/block.rs
+++ b/src/librustc_codegen_ssa/mir/block.rs
@@ -11,6 +11,7 @@ use crate::common::{self, IntPredicate};
 use crate::meth;
 
 use std::ffi::CString;
+use std::env;
 
 use crate::traits::*;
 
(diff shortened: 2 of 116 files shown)


vext01:
I think the plan of action is as follows:

 * Rename all instances of `no_trace`, to `no_sw_trace`.
 * Add a compile time switch which allows us to select which kind of tracing to use.
 * Compile-time guard all software tracing code so it only gets compiled in if we choose software tracing.
 * Make adjustments to this PR hooking guarding the hardware-tracing code in a similar fashion.
 * Merge this PR.

Sounds good?

The first item on the list is an easy one ;)

ltratt:
A dumb question: is the convention in rustc to use `XXX` or `FIXME`?

vext01:
You have to use `FIXME` or the tidy checker will get upset. It rejects `XXX` (which is unfortunate, as I prefer `XXX`).

vext01:
This is now ready for review.

On top of Lukas' label changes, we now have a unified `YK_TRACER` environment variable to (compile-time) choose which tracer to use.

In the long run (not now) we probably want this to be a proper (tracked) compiler switch, as this would cause cargo to rebuild stuff if the tracer choice changes.

ltratt:
Is it worth rebasing/squashing this before review?

vext01:
Ready for review.

src/librustc_codegen_llvm/builder.rs
```diff
@@ -154,12 +156,52 @@ impl BuilderMethods<'a, 'tcx> for Builder<'a, 'll, 'tcx> {
...
+    }
+
     fn position_at_end(&mut self, llbb: &'ll BasicBlock) {
         unsafe {
             llvm::LLVMPositionBuilderAtEnd(self.llbuilder, llbb);
         }
     }
 
+    fn add_yk_block_label(&mut self, di_sp: &DIScope, lbl_name: CString) {
+
```
ltratt: Blank line.

src/librustc_codegen_llvm/builder.rs
```diff
@@ -154,12 +156,52 @@ impl BuilderMethods<'a, 'tcx> for Builder<'a, 'll, 'tcx> {
...
+        // Insert a DWARF label at the start of each block.
+        // ykrt uses this at runtime to map virtual addresses to MIR blocks.
+        let first_instr = self.first_instruction(self.llbb());
+
+        if first_instr.is_some() {
```
ltratt:
```rust
if let Some(first_instr) = self.first_instruction(...) {
}
```

src/librustc_metadata/encoder.rs
```diff
@@ -65,6 +66,20 @@ macro_rules! encoder_methods {
     }
 }
 
+/// A check to see if we are using a Yorick tracer. Used to decide whether we need to encode MIR
+/// for all items.
+fn building_with_yk_tracer() -> bool {
+    match env::var("YK_TRACER") {
+        Ok(val) => {
+            match val.as_str() {
+                "hw" | "sw" | "sw-rustc" => true,
```
ltratt: Why do we have `sw` and `sw-rustc`?

src/librustc_codegen_ssa/mir/block.rs
```diff
@@ -803,7 +804,16 @@ impl<'a, 'tcx: 'a, Bx: BuilderMethods<'a, 'tcx>> FunctionCx<'a, 'tcx, Bx> {
 
         debug!("codegen_block({:?}={:?})", bb, data);
 
-        if bx.cx().has_debug() {
+        let tracer = match env::var("YK_TRACER") {
```
ltratt, line 807: That's the third time (I think) that this block of code appears somewhere. Should we find some way of putting it in a single place such that it can be called? I appreciate that this might not be easy.

src/librustc_metadata/encoder.rs
vext01: It's explained here:
https://github.com/softdevteam/ykrustc/pull/41/files#diff-1aeb4fd1a8c45243dc9f0a52c1dafc36R8

In short, it's a mode which inserts calls to the trace recorder, but which doesn't encode SIR. We use this for compiling ykrustc itself (we never want to trace the compiler binaries, but we do want to trace it's libraries).

ltratt: OK, got it.

src/librustc_codegen_ssa/mir/block.rs
vext01: There's a comment to this effect somewhere.

The *correct* thing to do is to have a compiler flag, e.g. a `-C tracer=X`. This has other advantages too: if we make the flag tracked, cargo will rebuild things if the flag changes.

Currently we just did the dumbest thing possible to get us going.

vext01: Here's the comment I mentioned:
https://github.com/softdevteam/ykrustc/pull/41/files#diff-24c5c945888bb0d041e769bfb852de6cR1105

ptersilie: I could maybe add a `FIXME` here...?

vext01: I don't think it's necessary to duplicate the existing `FIXME`.

I'm going to need to know whether we are going to use a flag or an environment for my CI tweaks though.

src/librustc_metadata/encoder.rs
vext01: In hindsight, I wonder if `sw-nosir` would have been a better name.

src/librustc_codegen_ssa/mir/block.rs
ltratt: Just to check: we're saying that because it's an environment variable we can't place this in a single function somewhere?

src/librustc_metadata/encoder.rs
ltratt: I tend to agree that's a better name. We can just add a commit at the end to that effect: I don't think we need to edit all the existing commits.

src/librustc_codegen_ssa/mir/block.rs
vext01: > Just to check: we're saying that because it's an environment variable we can't place this in a single function somewhere?

We can. I've started a branch to use a proper flag. It makes the code much nicer. We can cherry pick my commit into this PR once I'm done.

ltratt: Do you want to do that as a separate PR?

vext01: We can do if you like.

ltratt: I'm happy to do whatever's easiest for you. You choose.

vext01: Let's do it as a separate PR then.

ltratt: OK.

ltratt:
@ptersilie Please squash.

ptersilie:
Squashed and also added a commit renaming `sw-rustc` to `sw-nosir`.

vext01:
@ptersilie grep the whole tree for `sw-rustc`. There's at least one instance still present:
https://github.com/ptersilie/ykrustc/blob/yk_labels/.buildbot.sh#L12

ptersilie:
Oh, I wonder why `ag` didn't show me this!

ltratt:
Please squash.

ptersilie:
Squashed!

vext01:
@ptersilie ah I see you've met the tidy checker :)

```
tidy check
tidy error: /home/buildbot-worker8/worker8/buildbot-build-script/build/src/librustc_mir/transform/add_yk_swt_calls.rs:170: line longer than 100 chars
tidy error: /home/buildbot-worker8/worker8/buildbot-build-script/build/src/librustc_codegen_ssa/mir/block.rs:818: line longer than 100 chars
tidy error: /home/buildbot-worker8/worker8/buildbot-build-script/build/src/librustc_codegen_llvm/builder.rs:181: line longer than 100 chars
tidy error: /home/buildbot-worker8/worker8/buildbot-build-script/build/src/librustc_codegen_llvm/builder.rs:192: line longer than 100 chars
tidy error: /home/buildbot-worker8/worker8/buildbot-build-script/build/src/rustllvm/RustWrapper.cpp:958: line longer than 100 chars
```

vext01:
(you can invoke this by running `./x.py test` and waiting no longer than about 10 seconds)

ptersilie:
Ah, do I need to fix these manually? Or can I use clippy or something?

vext01:
It's manual. The compiler codebase doesn't use clippy or rustfmt.

ptersilie:
Try this.

vext01:
LGTM

ltratt:
Please squash.

ptersilie:
Squashed!

PR #42, opened 2019-09-11
Implement and use a query to help us find MIR in dependent crates.

vext01:
The new query returns a set of DefIds which:
 * Have an entry in the definition path table in the metadata.
 * Have MIR available.

At serialisation time, we then iterate over all dependent crates, adding
new DefIds into the list of things to be serialized.

I've not run the tests all the way yet, so we may encounter a couple of test failures. Can fix along the way.

(Following this merge, there will be a house-keeping PR where I clean up some old code we no-longer need).

### DIFF
--- a/src/librustc_interface/passes.rs
+++ b/src/librustc_interface/passes.rs
@@ -6,7 +6,7 @@ use log::{debug, info, warn, log_enabled};
 use rustc::dep_graph::DepGraph;
 use rustc::hir;
 use rustc::hir::lowering::lower_crate;
-use rustc::hir::def_id::{CrateNum, LOCAL_CRATE};
+use rustc::hir::def_id::{CrateNum, LOCAL_CRATE, DefId, DefIndex};
 use rustc::lint;
 use rustc::middle::{self, reachable, resolve_lifetime, stability};
 use rustc::middle::privacy::AccessLevels;
@@ -58,6 +58,7 @@ use serialize::json;
 use tempfile::Builder as TempFileBuilder;
 
 use std::any::Any;
+use std::borrow::Borrow;
 use std::env;
 use std::ffi::OsString;
 use std::fs;
@@ -1102,7 +1103,11 @@ pub fn start_codegen<'tcx>(
 
     // Output Yorick debug sections into binary targets.
     if tcx.sess.crate_types.borrow().contains(&config::CrateType::Executable) {
-        let (def_ids, _) = tcx.collect_and_partition_mono_items(LOCAL_CRATE);
+        let def_ids = tcx.collect_and_partition_mono_items(LOCAL_CRATE).0;
+        let mut def_ids = (*def_ids).clone();
+        for cnum in tcx.crates() {
+            def_ids.extend(tcx.defids_with_mir(*cnum));
+        }
         let sir_mode = if tcx.sess.opts.output_types.contains_key(&OutputType::YkSir) {
             // The user passed "--emit yk-sir" so we will output textual SIR and stop.
             SirMode::TextDump(outputs.path(OutputType::YkSir))
--- a/src/librustc_metadata/decoder.rs
+++ b/src/librustc_metadata/decoder.rs
@@ -11,6 +11,7 @@ use rustc::middle::exported_symbols::{ExportedSymbol, SymbolExportLevel};
 use rustc::hir::def::{self, Res, DefKind, CtorOf, CtorKind};
 use rustc::hir::def_id::{CrateNum, DefId, DefIndex, LocalDefId, CRATE_DEF_INDEX, LOCAL_CRATE};
 use rustc::hir::map::definitions::DefPathTable;
+use rustc::util::nodemap::DefIdSet;
 use rustc_data_structures::fingerprint::Fingerprint;
 use rustc::middle::lang_items;
 use rustc::mir::{self, interpret};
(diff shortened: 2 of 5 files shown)


src/librustc_interface/passes.rs
```diff
@@ -1102,7 +1103,12 @@ pub fn start_codegen<'tcx>(
 
     // Output Yorick debug sections into binary targets.
     if tcx.sess.crate_types.borrow().contains(&config::CrateType::Executable) {
-        let (def_ids, _) = tcx.collect_and_partition_mono_items(LOCAL_CRATE);
+        let def_ids = tcx.collect_and_partition_mono_items(LOCAL_CRATE).0;
```
vext01, line 1106: First let binding here is needed to keep a temporary alive long enough.

src/librustc_interface/passes.rs
```diff
@@ -1102,7 +1103,12 @@ pub fn start_codegen<'tcx>(
...
+        let def_ids: &DefIdSet = def_ids.borrow();
```
vext01: This binding is needed to avoid:
```
error[E0282]: type annotations needed
    --> src/librustc_interface/passes.rs:1107:54
     |
1107 |         let mut def_ids: DefIdSet = def_ids.borrow().clone::<DefIdSet>();
     |                                                      ^^^^^ cannot infer type for `Borrowed`
     |
     = note: type must be known at this point
```

I'd like to get rid of the intermediate binding if possible, but all of my attempts so far have failed.

Any ideas how to fix?

ltratt: I suspect there's an implicit `deref` or `as_ref` or similar being added to this so that this line really expands to something like:

```rust
let def_ids: &DefIdSet = def_ids.borrow().deref();
```

because otherwise the `Borrowed` type that `borrow()` returns wouldn't be compatible with `DefIdSet`.

src/librustc_metadata/decoder.rs
```diff
@@ -1147,6 +1148,22 @@ impl<'a, 'tcx> CrateMetadata {
         }
     }
 
+    pub fn defids_with_mir(&self, tcx: TyCtxt<'a, 'tcx, 'tcx>) -> DefIdSet {
+        if self.proc_macros.is_some() {
+            // Following the precedent of `exported_symbols()` we skip procedural macro crates.
+            DefIdSet::default()
+        } else {
+            let mut res = DefIdSet::default();
+            for idx in 1..self.root.entries_index.len - 1 {
```
ltratt, line 1157: Why do we skip `idx` 0?

ltratt:
bors try

src/librustc_metadata/decoder.rs
vext01: Index zero is the crate itself, which won't have any MIR attached. Although the call to `tcx.is_mir_available()` would also perform that filtering. I don't mind making it `0..`.

src/librustc_interface/passes.rs
vext01: So shall we leave it as is?

src/librustc_metadata/decoder.rs
ltratt: I would use whatever is most idiomatic: I have no opinion on this, but just wanted to flag it as something in case you hadn't fully thought about it.

src/librustc_interface/passes.rs
ltratt: Well, if you can change it to one line, it would probably be easier to read, and you'll know for next time, but it's up to you.

vext01: I'm confused because the set is behind an `Arc` whose borrow is implemented like this:
```
#[stable(feature = "rust1", since = "1.0.0")]
impl<T: ?Sized> borrow::Borrow<T> for Arc<T> {
    fn borrow(&self) -> &T {
        &**self
    }
}
```

The type parameter `Borrowed` is only visible in the trait def:
```
pub trait Borrow<Borrowed: ?Sized> {
    fn borrow(&self) -> &Borrowed;
```

ltratt: Ah, OK, then I don't know what the type inference error you encountered was!

vext01: Apparently it's a clash of the clones (sounds like a star wars film). Fix coming.

vext01:
Last push gets rid of one let binding. Let's see if the try passes before squash?

ltratt:
OK.

vext01:
> memory allocation of 224 bytes failed

We can try 12GiB heap again to get us going. I may be able to optimise things later.

vext01:
bors try

vext01:
So the build took 7 hours and 17 minutes. This is disappointing.

I think it's due to the sheer volume of stuff being serialised.

vext01:
[ignore travis btw -- that shouldn't show up next PR]

ltratt:
Please squash. So long as it works, the time isn't totally critical (though, obviously, it is a bit annoying).

vext01:
> though, obviously, it is a bit annoying

I'm really annoyed :(

Nonetheless we can move forward, but I'd like to get the turnover faster at some point.

vext01:
Squashed, and expanded the commit message quite a bit.

Please have a read, as I'd be interested to hear what you think of the compiler flag idea. I don't think it will be hard to implement.

ltratt:
Adding a compiler / environment flag to disable building this stuff during testing works well for me.

bors r+

PR #44, opened 2019-09-19
Replace YK_TRACER with a proper tracked compiler flag.

vext01:
See the commit message for a detailed description of this change.

We may even get a small CI speed up, as all ykrustc code that is *not* the standard library will not have calls to the trace recorder any more. As we discussed before, we never need to trace the compiler itself.

### DIFF
--- a/src/librustc/session/config.rs
+++ b/src/librustc/session/config.rs
@@ -99,6 +99,42 @@ pub enum LtoCli {
     Unspecified,
 }
 
+#[derive(Clone, Copy, PartialEq, Hash, Debug)]
+pub enum TracerMode {
+    /// `-C tracer=off`
+    Off,
+    /// `-C tracer=sw`
+    Software,
+    /// `-C tracer=hw`
+    Hardware,
+}
+
+impl TracerMode {
+    /// Does this mode require SIR to be encoded into the resulting binary?
+    pub fn encode_sir(&self) -> bool {
+        match self {
+            TracerMode::Off => false,
+            TracerMode::Software | TracerMode::Hardware => true,
+        }
+    }
+
+    /// Does this mode require DILabels for SIR mapping?
+    pub fn sir_labels(&self) -> bool {
+        match self {
+            TracerMode::Hardware => true,
+            TracerMode::Software | TracerMode::Off => false,
+        }
+    }
+
+    /// Does this mode require each block to call the software trace recorder?
+    pub fn sw_trace_recorder(&self) -> bool {
+        match self {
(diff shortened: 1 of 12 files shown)


src/librustc/session/config.rs
```diff
@@ -1095,6 +1133,16 @@ macro_rules! options {
             true
         }
 
+        fn parse_tracer(slot: &mut TracerMode, v: Option<&str>) -> bool {
+            *slot = match v {
+                None | Some("off") => TracerMode::Off,
+                Some("sw") => TracerMode::Software,
+                Some("hw") => TracerMode::Hardware,
+                Some(_) => return false,
```
ltratt, line 1141: Should we be `panic`ing if we get passed something we don't know how to deal with?

vext01: Returning false tells the compiler to give an appropriate error message:
```
$ rustc -C tracer=taps src/main.rs 
error: incorrect value `taps` for codegen option `tracer` - one of `off`, `sw`, `sw-nosir` or `hw` was expected
```

But we do have a bug in the message! `sw-nosir` is gone! Fixing.

ltratt: OK.

vext01: Fixed

ltratt:
If @ptersilie is OK with this, then you should squash.

src/librustc/session/config.rs
```diff
@@ -846,6 +882,8 @@ macro_rules! options {
         pub const parse_lto: Option<&str> =
             Some("either a boolean (`yes`, `no`, `on`, `off`, etc), `thin`, \
                   `fat`, or omitted");
+        pub const parse_tracer: Option<&str> =
+            Some("one of `off`, `sw`, `sw-nosir` or `hw`");
```
ptersilie: You said we removed `sw-nosir`, so I believe this one should go as well.

vext01: Isn't that the same one?

vext01: Just grepped the tree to be sure. No instances of `sw-nosir` remain.

ptersilie: Ah, you fixed this already in a later commit? I hadn't gotten so far yet.

ptersilie:
Yes, looks good. Please squash.

vext01:
splat

vext01:
Sorry, my bad. Tidy check. OK to squash?

ptersilie:
Yes!

vext01:
splat

PR #49, opened 2019-09-23
Parallel CI

vext01:
~~Testing. Please do not review.~~

This should now work. I'm interested to see how long the build takes when running in parallel.

If all goes to plan, I'll have another PR for the bulidbot config repo.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -5,9 +5,6 @@
 set -e
 
 export PATH=PATH=/opt/gdb-8.2/bin:${PATH}
-# Select which kind of traceable std lib to build.
-# FIXME eventually CI will test both modes and thus set this itself.
-export STD_TRACER_MODE=sw
 
 TARBALL_TOPDIR=`pwd`/build/ykrustc-stage2-latest
 TARBALL_NAME=ykrustc-stage2-latest.tar.bz2
--- a/bors.toml
+++ b/bors.toml
@@ -1,6 +1,7 @@
 # The service providing the commit statuses to GitHub.
 status = [
-  "buildbot/buildbot-build-script"
+  "buildbot/ykrustc-buildbot-build-script-hwt",
+  "buildbot/ykrustc-buildbot-build-script-swt",
 ]
 
 # Allow eight hours for builds + tests.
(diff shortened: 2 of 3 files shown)


vext01:
bors try

vext01:
To clarify, this is now ready for review.

vext01:
We can start bors on this now.

bjorn3:
The build log link gives a connection reinitialized error.

vext01:
```
---- net::tcp::tests::fast_rebind stdout ----
thread '<unnamed>' panicked at 'received error for `TcpListener::bind(&addr)`: Address already in use (os error 98)', src/libstd/net/tcp.rs:1283:13
```

This is most unfortunate. Could it be that one of the builds binds an address and the other build also tries to, but fails?

vext01:
Possible solutions:
 * Add a sleep to the start to one of the builds so as to attempt to avoid test clashing.
 * Retry the tests in a loop (already passing suites should be skipped)
 * A proper fix

I think this is worthy of an upstream bug report nonetheless.

ltratt:
> Could it be that one of the builds binds an address and the other build also tries to, but fails?

Sounds plausible. I think the onus is on us to fix/update the test, however.

If you look at https://github.com/rust-lang/rust/blob/master/src/libstd/net/test.rs#L35 (the `base_port`) function, you can see that there's an attempt to avoid this problem, but it doesn't seem to fully work. I think that's where I'd focus our efforts on, at least at first.

vext01:
> If you look at...

Ah, nice find. Yes, I see why this doesn't work for us. Watch this space.

vext01:
That last push should fix it.

OK?

ltratt:
Fine with me. Can you reword the commit title to remove the extraneous "for"? Then I can start bors.

ltratt:
Fine with me. Can you reword the commit title to remove the extraneous "for"? Then I can start bors.

vext01:
Oops. It was supposed to be "from" not "for". Fixed.

ltratt:
I wonder why we got a timeout and then success?

bjorn3:
I think that timeout is for the try from https://github.com/softdevteam/ykrustc/pull/49#issuecomment-534021015 which got issued 9 hours ago.

ltratt:
That had not occurred to me as a possibility -- good spot!

PR #50, opened 2019-09-24
Build separate tarballs for hardware- and software-enabled ykrustc.

vext01:
This makes the CI for ykrustc build two separate tarballs for each tracer mode.

Eventually the `yk` repo will have parallel CI which will pick up these two tarballs.

I suggest we wait until we've made the yksom CI changes before kicking bors on this one. When the time comes, we can do a `try` and I can check the local filesystem on the CI server. If it works, we can `r+` it.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -4,10 +4,16 @@
 
 set -e
 
+case ${STD_TRACER_MODE} in
+    "sw" | "hw") true;;
+    *) echo "STD_TRACER_MODE must be set to either 'hw' or 'sw'"
+       exit 1;;
+esac
+
 export PATH=PATH=/opt/gdb-8.2/bin:${PATH}
 
 TARBALL_TOPDIR=`pwd`/build/ykrustc-stage2-latest
-TARBALL_NAME=ykrustc-stage2-latest.tar.bz2
+TARBALL_NAME=ykrustc-${STD_TRACER_MODE}-stage2-latest.tar.bz2
 SNAP_DIR=/opt/ykrustc-bin-snapshots
 
 # Ensure the build fails if it uses excessive amounts of memory.


bjorn3:
Is the difference between the two only in which mode libstd is build? In that case it would be possible to build rustc once and then build separate libstd with -Ctracer=sw and -Ctracer=hw.

vext01:
@bjorn3 That's right. But it wasn't clear to me if one rustc could load another's std and core etc. For example, I notice that the compiler rlibs have some kind of hash or fingerprint appended to their filename, and I'm not sure if that may cause any issue.

We may revisit this later, but top priority now is to do something simple to get a minimal working system.

bjorn3:
> But it wasn't clear to me if one rustc could load another's std and core etc. 

It is possible as long as it is compiled using that rustc. Just pass `--sysroot` to a dir with the right layout.

> For example, I notice that the compiler rlibs have some kind of hash or fingerprint appended to their filename, and I'm not sure if that may cause any issue.

No thats not a problem. As long as there is just one libstd-*.rlib, ... in the sysroot there is no problem.

https://github.com/japaric/xargo is useful. Despite being in maintenance mode, it is kept up to date with the latest rustc.

vext01:
Very cool. Thanks for pointing this out. We will probably revisit this in the future!

.buildbot.sh
```diff
@@ -4,10 +4,16 @@
 
 set -e
 
+case ${STD_TRACER_MODE} in
+    "sw" | "hw") true;;
+    *) echo "STD_TRACER_MODE must be set to either 'hw' or 'sw'"
+        exit 1;;
```
ltratt: Minor inconsistent indent?

vext01: Yeah, I was unsure how to format this. How would you have done it?

ltratt: I would line the `exit` up to `echo` myself.

vext01:
Ready to squash?

ltratt:
Please squash.

vext01:
splat.

PR #54, opened 2019-09-26
Remove the work loop from the SIR encoding code.

vext01:
This change removes the work list from the SIR encoding section of ykrustc. We used to try and find things to encode by following calls in the MIR, but now we have a different way of doing this.

Assigning to @ptersilie as @ltratt is away today.


ptersilie:
Looks good, but maybe we can expand on the commit message a bit which is currently a bit vague, e.g. what is no longer needed, and what are `other means`?

vext01:
Give this a go.

PR #59, opened 2019-10-04
Add YK labels when returning from a function call

ptersilie:
The MIR level and the binary level handle function calls that return a
value differently. The former treats the assignment of that value as
part of the function block, whereas the latter treats it as part of the
block following the return. Although, this doesn't make a difference
when running programs, it means that the block label for the following
block is at the wrong location, which makes it impossible to assign it
to the correct trace.

However, we can't simply adjust the label by x bytes, since this problem
only occurs when the result of the function call is assigned to a
variable. We thus add another label `YK_RET` which is inserted during a
function call, just before the return value is assigned. This places
this new label at the correct location and allows us to match it to the
same location in our trace.

### DIFF
--- a/src/librustc_codegen_ssa/mir/block.rs
+++ b/src/librustc_codegen_ssa/mir/block.rs
@@ -143,6 +143,15 @@ impl<'a, 'tcx> TerminatorCodegenHelper<'a, 'tcx> {
             }
 
             if let Some((ret_dest, target)) = destination {
+                // Generate YK debug label to match hardware traces with blocks in the MIR.
+                if bx.cx().tcx().sess.opts.cg.tracer.sir_labels() && bx.cx().has_debug() {
+                    let did = fx.instance.def.def_id();
+                    let lbl_name = CString::new(format!("__YK_RET_{}_{}_{}", did.krate.as_u32(),
+                                                did.index.as_u32(), self.bb.index())).unwrap();
+                    let di_sp = fx.fn_metadata(self.terminator.source_info.span);
+                    bx.add_yk_block_label_at_end(*di_sp, lbl_name);
+                }
+
                 fx.store_return(bx, ret_dest, &fn_ty.ret, llret);
                 self.funclet_br(fx, bx, target);
             } else {


ptersilie:
@vext01 Maybe have a good look at the PR/Commit description too. I tried to explain it as best as I could, but may have gotten some of the details wrong.

src/librustc_codegen_ssa/mir/block.rs
```diff
@@ -143,6 +143,15 @@ impl<'a, 'tcx> TerminatorCodegenHelper<'a, 'tcx> {
             }
 
             if let Some((ret_dest, target)) = destination {
+                // Generate debug label
```
vext01: Can we expand this comment, making it clear that this is to do with Yorick, and not regular debugging stuff?

src/librustc_codegen_ssa/mir/block.rs
```diff
@@ -143,6 +143,15 @@ impl<'a, 'tcx> TerminatorCodegenHelper<'a, 'tcx> {
...
+                if bx.cx().tcx().sess.opts.cg.tracer.sir_labels() && bx.cx().has_debug() {
+                    let did = fx.instance.def.def_id();
+                    let lbl_name = CString::new(format!("__YK_RET_{}_{}_{}", did.krate.as_u32(),
```
vext01, line 149: At some point, all of our labels should use a crate hash (`tcx.crate_hash(did)`) instead of a crate number. This can be done in a future PR.

The problem with crate numbers is that they change depending upon which dependency is being built, whereas crate hashes are stable over the entire compilation session.

vext01:
Couple of comments, but mostly LGTM.

ptersilie:
Did you check the PR/commit message?

vext01:
Yes, it seems fine to me.

ptersilie:
See if that works for you.

vext01:
LGTM, please squash.

ptersilie:
Squashed!

vext01:
Branch needs sync

ptersilie:
Rebased.

vext01:
Not sure what happened here. Let's retry.

bors r+

vext01:
Ah, this r+ is queued. Another ykrustc build is already happening.

PR #30, opened 2019-06-24
TIR -> SIR.

vext01:
Companion PR to https://github.com/softdevteam/yk/pull/15.

PR #64, opened 2019-10-28
Complete cycle breaker for #30

vext01:
Also update lock file.


vext01:
```
---- net::tcp::tests::fast_rebind stdout ----
thread '<unnamed>' panicked at 'received error for `TcpListener::bind(&addr)`: Address already in use (os error 98)', src/libstd/net/tcp.rs:1283:13
```

The two builds clashed. I thought we had fixed this...

vext01:
Here's the discussion from an earlier PR where we fixed this:
https://github.com/softdevteam/ykrustc/pull/49#issuecomment-534105680

It's possible that one of us was building manually at the same time as CI was running...

We could just retry.

PR #69, opened 2019-11-05
Replace crate ids in YK labels with crate hashes

ptersilie:
In order to map labels back to SIR blocks, we need the crate hash.
Currently we are only putting the crate id into the label, so we need to
replace it with the crate hash.


vext01:
LGTM, but can we explain that the crate number is not stable across dependencies, whereas the crate hash is (in the commit message).

ptersilie:
See if that new commit message is better.

vext01:
Please squash, but replace "crate ids" with "crate numbers" in the message, as that's what rustc calls them (it's the `CrateNum` type).

ptersilie:
Fixed.

PR #71, opened 2019-11-11
Don't generate labels if there is no DIScope

ptersilie:
Recent changes in rustc made `debug_context` an `Option` and there seems to be no guarantee that we will always have one. For example, this caused `rustc` to segfault when running `cargo test` (even on an empty project) with hardware tracing enabled.

Since there is nothing we can do if there is no debug context, we simply disable labels which means we won't be able to trace that particular function/block. However, since this should happen very rarely, we should barely notice this.

### DIFF
--- a/src/librustc_codegen_ssa/mir/block.rs
+++ b/src/librustc_codegen_ssa/mir/block.rs
@@ -149,8 +149,9 @@ impl<'a, 'tcx> TerminatorCodegenHelper<'a, 'tcx> {
                     let crate_hash = bx.cx().tcx().crate_hash(did.krate).as_u64();
                     let lbl_name = CString::new(format!("__YK_RET_{}_{}_{}", crate_hash,
                                                 did.index.as_u32(), self.bb.index())).unwrap();
-                    let di_sp = fx.fn_metadata(self.terminator.source_info);
-                    bx.add_yk_block_label_at_end(di_sp, lbl_name);
+                    if let Some(di_sp) = fx.fn_metadata(self.terminator.source_info) {
+                        bx.add_yk_block_label_at_end(di_sp, lbl_name);
+                    }
                 }
 
                 fx.store_return(bx, ret_dest, &fn_abi.ret, llret);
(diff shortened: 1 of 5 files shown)


src/librustc_codegen_ssa/mir/block.rs
```diff
@@ -150,7 +150,9 @@ impl<'a, 'tcx> TerminatorCodegenHelper<'a, 'tcx> {
                     let lbl_name = CString::new(format!("__YK_RET_{}_{}_{}", crate_hash,
                                                 did.index.as_u32(), self.bb.index())).unwrap();
                     let di_sp = fx.fn_metadata(self.terminator.source_info);
-                    bx.add_yk_block_label_at_end(di_sp, lbl_name);
+                    if di_sp.is_some() {
```
vext01: How about:

```
if let Some(di_sp) = di_sp {
   ...
}
```

Avoids an unwrap.

vext01:
This LGTM. One small comment.

ptersilie:
Fixed. We still need to add the `cargo test` test somehow, so we won't break this again in the future.

vext01:
> Fixed. We still need to add the cargo test test somehow, so we won't break this again in the future.

Perhaps you can add a compile-pass test, passing the option you found the other day to build the tests (`--test`?). Failing that, you can add another run-make test

vext01:
> Perhaps you can add a compile-pass test, passing the option you found the other day to build the tests (--test?)

Or did that not crash? I can't remember.

ptersilie:
`--test` did not crash, no.

vext01:
`run-make` it is then.

ptersilie:
bors try

ptersilie:
bors try

ptersilie:
bors try

ptersilie:
Finally. Ready for review and squashing.

vext01:
Please remove your email address and force push.

ptersilie:
Squashed!

PR #73, opened 2019-11-13
Add a tracermode attribute

ptersilie:
This allows us to deactivate specific tests depending on which
tracermode is currently enabled.

### DIFF
--- a/src/librustc/session/config.rs
+++ b/src/librustc/session/config.rs
@@ -1591,6 +1591,17 @@ pub fn default_configuration(sess: &Session) -> ast::CrateConfig {
     if sess.opts.crate_types.contains(&CrateType::ProcMacro) {
         ret.insert((sym::proc_macro, None));
     }
+    match sess.opts.cg.tracer {
+        TracerMode::Software => {
+            ret.insert((Symbol::intern("tracermode"), Some(Symbol::intern("sw"))));
+        },
+        TracerMode::Hardware => {
+            ret.insert((Symbol::intern("tracermode"), Some(Symbol::intern("hw"))));
+        },
+        TracerMode::Off => {
+            ret.insert((Symbol::intern("tracermode"), Some(Symbol::intern("off"))));
+        }
+    }
     ret
 }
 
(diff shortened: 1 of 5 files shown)


vext01:
This looks fine.

Did you get anywhere with the test we discussed yesterday?

src/librustc/session/config.rs
```diff
@@ -1591,6 +1591,17 @@ pub fn default_configuration(sess: &Session) -> ast::CrateConfig {
     if sess.opts.crate_types.contains(&CrateType::ProcMacro) {
         ret.insert((sym::proc_macro, None));
     }
+    match sess.opts.cg.tracer {
+        TracerMode::Software => {
+            ret.insert((Symbol::intern("tracermode"), Some(Symbol::intern("sw"))));
```
bjorn3, line 1596: Maybe prefix with `yk_`?

ptersilie: I tried to make it similar to the command line arguments that invoke the tracer, e.g. `-C tracer=hw`. So if we were to add `yk_` we should add that to the compiler argument as well. Alternatively, maybe we should also rename `tracermode` to `tracer` so it's the same as the compiler flag?

ptersilie:
I forgot about the test. Will add try and add one soon.

ptersilie:
@vext01 These tests should do the trick. It didn't work the way we discussed yesterday, so this solution is slightly more convoluted than what we had before.

src/test/run-make/yk-tracer-attrs/main.rs
```diff
@@ -0,0 +1,4 @@
+fn main() {
```
vext01: Is this file used?

vext01:
Looks fine, but you may have added one more file than necessary?

src/test/run-make/yk-tracer-attrs/main.rs
ptersilie: Whoops. Yes, this needs to go.

src/librustc/session/config.rs
vext01: We'd probably also have to change `#[no_swt]`. Let's roll without the prefix for now, and if things get confusing at any point, we can add the prefix.

Thanks for the suggestion Bjorn!

vext01:
Are we ready for squashing?

ptersilie:
Squashed!

PR #80, opened 2020-01-13
Start moving SIR generation to LLVM codegen.

vext01:
Previously I'd tried implementing SIR generation as a new codegen from
scratch. What I learned from this is:

 - There's a lot of boilerplate in the codegen API.

 - To get it right, you'd end up duplicating a lot of LLVM backend code,
   which would need to be kept in sync over time. This code often
   contains low-level details, which at this point in time I don't
   understand. The code is also hard to debug if you get it wrong.

 - Pretty much the whole codegen has to be implemented and working all
   at once. You cannot really incrementally implement parts.

I spent quite some time trying that approach. I generated about 2000
lines of backend code. It's slow going and relentless.

Due to frustration with all of the above, I took pause to see if I could
find a better way.

This change shows an idea I had at the end of last week. The general
idea is to -- rather than implement a whole new backend -- add calls to
SIR generation from within the existing LLVM backend. This means that:

 - We can build our SIR IR incrementally over time.

 - We don't have to understand all of the low-level details right off
   the bat.

 - There will be no code duplication and when the LLVM backend changes,
   instead of our code going out of sync we will get merge conflicts
   (in a good way).

 - The code should be smaller.

This all sounds good, but to make it work, there's one small detail. The
LLVM codegen is opaque-pointer-centric and such pointers are often used
as identifiers in LLVM API calls. For example you "add a new block to the
function behind the following opaque Value pointer". To generate SIR for
such a scenario we have to keep a mapping of LLVM pointers to the
corresponding SIR. For example so that we know which of the SIR
functions to add a block to. This will raise memory consumption a bit,
but I hope it isn't all that bad.

To demonstrate the approach, the following diff shows creation of SIR functions and blocks in-memory. Eventually we'd be serialising this someplace for picking up at runtime.

I'm raising a draft PR for comments and discussion before I proceed, as it's a bold deviation from what I've been doing previously.

Thoughts?


ltratt:
You know far more about this than I do: if you think this is the right route to try then I will support you. Certainly, your description makes it seem like a reasonable route to try!

vext01:
> Certainly, your description makes it seem like a reasonable route to try

Cool. Let's give it a shot.

On top of the change in this PR, I'm going to work on getting the IR serialized, then we can do incremental PRs.

ltratt:
Should I bors this?

vext01:
> Should I bors this?

Not just yet. I want to check I can get serialised objects into executables and rlibs first.

(So far so good)

ptersilie:
This looks good to me as well. I can actually sort-of understand what's going on, which means this is probably the right approach.

vext01:
Finally all of the tests are passing and we can start reviewing our changes.

I apologise profusely for inflicting this on you @ltratt -- There's no denying that it's going to be horrible to review. To be honest, I didn't realise it would be such a large change.

The plan would be too splat this all into one commit, as trying to get it into testable subsets is going to be too hard.

I'm also going to point out a bit of code via a comment.

src/librustc/sir.rs
```diff
@@ -0,0 +1,136 @@
...
+        debug_assert!(existing.is_none());
+    }
+
+    pub fn get_symbol_name(&mut self, func: *const Value) -> &String {
+        let func_idx = self.llvm_values[&func].func_idx();
+        let sir_func = &mut self.funcs[func_idx];
+        &sir_func.symbol_name
+    }
+
+    /// For hardware tracing, during codegen we insert DILabels to know where we are in the binary.
```
vext01, line 98: This bit of code *probably* isn't required, as no reproducible build test actually uses `-C tracer` at present.

 I implemented this because I thought it was causing a test failure, but the real problem was elsewhere: we were erroneously creating and linking empty LLVM SIR modules when `-C tracer` is **not** used. (The SIR objects we link have non-deterministic paths).

This code may as well stay, as it doesn't noticeably slow us down and we might one day want to test reproducible builds with tracing enabled. I know how to fix it, just it's not top priority right now.

src/librustc/sir.rs
```diff
@@ -0,0 +1,136 @@
...
+    pub fn func_idx(&self) -> SirFuncIdx {
+        let Self::Func(idx) = self;
+        *idx
+    }
+}
+
+pub struct SirCx {
+    /// Maps an opaque LLVM `Value` to its SIR equivalent.
+    pub llvm_values: FxHashMap<*const Value, SirValue>,
+    /// Maps an opaque LLVM `BasicBlock` to the function and block index of it's SIR equivalent.
```
ltratt: "it's" -> "its"

src/librustc/sir.rs
```diff
@@ -0,0 +1,136 @@
...
+        }
+    }
+
+    pub fn add_func(&mut self, value: *const Value, symbol_name: String) {
+        let idx = SirFuncIdx::from_usize(self.funcs.len());
+
+        self.funcs.push(ykpack::Body{
+            symbol_name,
+            blocks: Default::default(),
+            num_locals: 0,  // FIXME
```
ltratt: Can we make these `assert`s on something? I always find `FIXME` dangerous, but I know it's occasionally unavoidable.

src/librustc/sir.rs
```diff
@@ -0,0 +1,136 @@
...
+pub struct SirCx {
+    /// Maps an opaque LLVM `Value` to its SIR equivalent.
+    pub llvm_values: FxHashMap<*const Value, SirValue>,
+    /// Maps an opaque LLVM `BasicBlock` to the function and block index of it's SIR equivalent.
+    pub llvm_blocks: FxHashMap<*const BasicBlock, (SirFuncIdx, SirBlockIdx)>,
+    /// Function store. Also owns the blocks
+    pub funcs: IndexVec<SirFuncIdx, ykpack::Body>,
+}
+
+impl SirCx {
```
ltratt, line 56: Do any of the functions in this `impl` have any non-obvious constraints that need documenting? `add_func` sounds pleasingly obvious but maybe, I dunno, "the user must ensure no otehr function with this name exists" or whatever?

src/librustc/sir.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /// For hardware tracing, during codegen we insert DILabels to know where we are in the binary.
+    /// These labels must be emitted in a deterministic order otherwise the reproducible build
+    /// checker gets upset. This function gives the codegen what it needs in a data structure which
+    /// can be iterated deterministically.
+    pub fn funcs_and_blocks_deterministic(&self)
+        -> IndexVec<SirFuncIdx, IndexVec<SirBlockIdx, *const BasicBlock>>
+    {
+        // Scaffold a data structure which can be iterated deterministically by the caller.
```
ltratt: "Scaffold" is one of those words that I find semantically meaningless. Can we use something clearer?

src/librustc/sir.rs
```diff
@@ -0,0 +1,136 @@
...
+        // To start, all pointer addresses are None. We would have use overwritable null pointers,
```
ltratt: "we would have TO use"?

ltratt: I also don't understand the `None` / "overwriteable null pointers" comparison.

src/librustc/sir.rs
```diff
@@ -0,0 +1,136 @@
...
+            res[func_idx] = IndexVec::from_elem_n(None, func.blocks.len());
+        }
+
+        // Now populate the pointers.
+        for (bb, (func_idx, bb_idx)) in &self.llvm_blocks {
+            debug_assert!(res[*func_idx][*bb_idx] == None);
+            res[*func_idx][*bb_idx] = Some(*bb);
+        }
+
+        // Now get rid of the Option wrappers around our pointers. There should be no Nones left.
```
ltratt: The "There should be no `None`s left" seems redundant: `b.unwrap()` guarantees that, as does the type system given the return type for `ret`.

src/librustc/sir.rs
```diff
@@ -0,0 +1,136 @@
...
+        let mut ret = IndexVec::default();
+        for func_blocks in res.into_iter() {
+            ret.push(func_blocks.into_iter().map(|b| b.unwrap()).collect());
+        }
+
+        ret
+    }
+
+    /// Dump SIR to text file.
+    /// Used in tests and for debugging.
```
ltratt, line 128: Should this be `#[test]`?

src/librustc_codegen_llvm/base.rs
```diff
@@ -168,10 +170,58 @@ pub fn compile_codegen_unit(
...
+
+                        // When --test is passed, the auto-generated entry point causes our label
+                        // insertion to crash. We never want to trace test harnesses anyway, so
+                        // skip them.
+                        if tcx.sess.opts.test {
+                            let entry_fn = tcx.entry_fn(LOCAL_CRATE).unwrap().0;
+                            let entry_inst = Instance::mono(tcx, entry_fn);
+                            let entry_sym = &*tcx.symbol_name(entry_inst).name.as_str();
+                            if sym_name == entry_sym {
+                                dbg!("skipping a test harness main");
```
ltratt: Is this `dbg` useful?

src/librustc_codegen_llvm/base.rs
```diff
@@ -168,10 +170,58 @@ pub fn compile_codegen_unit(
...
+                                continue;
+                            }
+                        }
+
+                        for (bb_idx, bb) in blocks.iter_enumerated() {
+                            let llbb = unsafe {
+                                &*(*bb as *const sir::BasicBlock as *const llvm::BasicBlock)
+                            };
+
+                            let lbl_name = CString::new(format!("__YK_BLK:{}:{}",
```
ltratt: Should the `__YK_BLK` be a constant somewhere? I guess we use this in more than one place?

src/librustc_codegen_llvm/context.rs
```diff
@@ -290,6 +294,17 @@ impl<'ll, 'tcx> CodegenCx<'ll, 'tcx> {
 
         let isize_ty = Type::ix_llcx(llcx, tcx.data_layout.pointer_size.bits());
 
+        // If we are putting SIR into the binary or dumping it to disk, then create a SirCx. Also
+        // skip generating SIR for build scripts (we will never trace them).
+        let sir_cx = if (tcx.sess.opts.cg.tracer.encode_sir() ||
+            tcx.sess.opts.output_types.contains_key(&OutputType::YkSir)) &&
+            tcx.crate_name(hir::def_id::LOCAL_CRATE).as_str() != "build_script_build"
```
ltratt, line 301: Is `build_script_build` guaranteed not to clash with some unfortunately named user crate?

src/librustc_codegen_llvm/sir.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    encoder.done().unwrap();
+
+    let (sir_llcx, sir_llmod) = (&*sir_llvm_module.llcx, sir_llvm_module.llmod());
+    let llmeta = common::bytes_in_context(sir_llcx, &buf);
+    let llconst = common::struct_in_context(sir_llcx, &[llmeta], false);
+
+    // Borrowed from exported_symbols::metadata_symbol_name().
+    let sym_name = format!("yksir_{}_{}",
```
ltratt: I wonder if the `yksir_` prefix should be a constant in case we want to change it?

src/librustc_codegen_ssa/back/link.rs
```diff
@@ -542,10 +552,17 @@ fn link_natively<'a, B: ArchiveBuilder<'a>>(sess: &'a Session,
...
-        cmd.arg("-Wl,--no-gc-sections");
-        cmd.args(sess.yk_link_objects.borrow().iter().map(|o| o.path()));
+    // Link Yorick SIR into executables.
+    // This links only the code from the local crate. See add_upstream_rust_crates() for where SIR
+    // for dependencies is linked.
+    //
+    // FIXME When rebuilding after a small change, only the rebuilt code gets SIR. We should cache
+    // the SIR in a file between builds and load the SIR from last time as a starting point, only
+    // updating what changed. This may require dumping the SIR into a file between builds, like
+    // what Rust does with metadata.
```
ltratt: "like what" -> "as"

src/librustc_codegen_ssa/back/link.rs
```diff
@@ -1121,7 +1138,9 @@ fn link_args<'a, B: ArchiveBuilder<'a>>(cmd: &mut dyn Linker,
 
     // Try to strip as much out of the generated object by removing unused
     // sections if possible. See more comments in linker.rs
-    if !sess.opts.cg.link_dead_code {
+    //
+    // For Yorick, we don't GC sections so as to keep SIR.
```
ltratt: Huh?

src/librustc_codegen_ssa/back/link.rs
```diff
@@ -1434,6 +1457,36 @@ fn add_upstream_rust_crates<'a, B: ArchiveBuilder<'a>>(
         add_static_crate::<B>(cmd, sess, codegen_results, tmpdir, crate_type, cnum);
     }
 
+    /// Extracts the SIR from the dependency rlib and puts it into it's own static object for
+    /// --whole-archive linkage. This is required because the link editor will normally only link
+    /// strictly required symbols and it is unaware that we will look directly into the SIR section
+    /// at runtime. Whilst we could link the originall rlib with --whole-archive, that would bloat
```
ltratt: "original" (one too many "l"s).

src/rustllvm/RustWrapper.cpp
```diff
@@ -985,24 +985,35 @@ extern "C" int64_t LLVMRustDIBuilderCreateOpPlusUconst() {
   return dwarf::DW_OP_plus_uconst;
 }
 
+// Adds YK labels as close to the beginning of BasicBlocks as possible.
+//
+// This is tricky because LLVM has certain constraints which the bitcode
+// checker will complain about. For example, you may not insert debug info
+// before PHI nodes.
```
ltratt, line 992: Are there any others?

src/rustllvm/RustWrapper.cpp
```diff
@@ -985,24 +985,35 @@ extern "C" int64_t LLVMRustDIBuilderCreateOpPlusUconst() {
   return dwarf::DW_OP_plus_uconst;
 }
 
+// Adds YK labels as close to the beginning of BasicBlocks as possible.
```
ltratt, line 988: Should this be a `///` doc comment?

ltratt:
My comments are all very shallow, as I don't understand most of this. That isn't a complaint: given all the work you've both done, I didn't expect to understand all the details! But please bear in mind that my comments are thus coming from a point of genuine ignorance.

src/librustc/sir.rs
vext01: I'm not sure what the assertion would be or if we even need these fields in the new SIR. I only included them so as to minimise changes to ykpack.

src/librustc/sir.rs
vext01: Ah yes. There is a constraint, but I'll have to look up the rules. We inherit some oddity from LLVM as I recall.

src/librustc/sir.rs
ltratt: If we don't know if we need them, wouldn't it be better to bite the pullet re: ykpack? Otehrwise I can imagine no-one will ever be brave enough to remove them!

src/librustc/sir.rs
vext01: The "scaffolding" here builds a data structure with mutable place-holder `None`s which are updated to `Some`s in the loop later. Rather than using `Option<*const ...>` and stripping the resulting `Some`s later, I'd have liked to have used `null` as the placeholder, but Rust refuses.

ltratt: OK. Can we maybe just tweak the comment a bit then?

src/librustc_codegen_llvm/base.rs
ptersilie: Currently we only use this here since moving label generation closer to LLVM. And I don't think we'll be adding blocks anywhere else. That being said, I'm not generally opposed to putting this in a constant.

src/librustc/sir.rs
vext01: eh? unwrap() doesn't guarantee anything. It can crash if there is a None.

You can think of this comment as one of your "unwrap() justifications" that you do in your PRs. If there was a None, then we did something wrong in the above loop.

src/librustc/sir.rs
vext01: No, because `--emit yk-sir` uses it regardless. We just have a test which uses this flag (at runtime).

src/librustc_codegen_llvm/base.rs
vext01: I was supposed to remove it. My bad.

src/librustc_codegen_llvm/base.rs
vext01: @ptersilie

src/librustc_codegen_llvm/context.rs
vext01: no, but that's the name the compiler assigns a build script, so it's kind of out of our control.

src/librustc_codegen_llvm/sir.rs
vext01: Can do, yeah. We can probably put that in ykpack.

src/rustllvm/RustWrapper.cpp
vext01: @ptersilie do you remember?

src/librustc/sir.rs
ltratt: Ah you meant there should be no `None`s *before* the `unwrap`! The comment suggested *after* the `unwrap` to me. Maybe tweak the phrasing?

src/librustc/sir.rs
ltratt: OK.

src/librustc_codegen_ssa/back/link.rs
vext01: This is a linking detail. Because the linker deems our SIR section unreferenced (it can't know we will load the binary from disk at runtime), it will remove it unless we tell it not to.

src/librustc_codegen_llvm/base.rs
ltratt: Empty comment from @vext01?

src/librustc_codegen_llvm/sir.rs
ltratt: OK.

src/librustc_codegen_ssa/back/link.rs
vext01: See also `--gc-sections` in the man page for `ld`.

src/rustllvm/RustWrapper.cpp
vext01: Note it's C++.

src/librustc_codegen_ssa/back/link.rs
ltratt: We probably need to make clearer that this is about linking? Should we also point readers at the `ld` man page for `--gc-sections`? [I have no idea about this stuff.]

src/rustllvm/RustWrapper.cpp
ltratt: Yes, but you can still add doc comments to it.

vext01:
> My comments are all very shallow

No worries. I think you've still found some good stuff.

I still feel pretty guilty for dumping this on you :P

src/librustc/sir.rs
vext01: I suppose it doesn't hurt. We can always bring it back later.

ltratt:
No need to feel guilty. This PR is clearly a result of a lot of hard, deep work!

src/librustc_codegen_llvm/base.rs
vext01: I was prompting @ptersilie to answer, because I wasn't sure :P

vext01: @ptersilie the mapper in `yk` will need this constant too, so perhaps it should be a constant in yk? I can do that if you like.

src/librustc_codegen_ssa/back/link.rs
vext01: There's a couple of places in upstream rust where they talk about some of these details, but I suppose it won't hurt to mention it here.

ltratt: Maybe we just need to point people at where to look upstream then?

src/rustllvm/RustWrapper.cpp
vext01: I'm happy to be proven wrong but I don't think `///` means anything to C++ and rustdoc doesn't look at C++ files.

The closesnt markup to `///` would be for doxygen:
```
/**
 * blah...
 */
```

But we don't use doxygen.

ltratt: I'm an idiot -- ignore me!

src/librustc/sir.rs
vext01: Here's what I was thinking of:
https://github.com/vext01/ykrustc/blob/083637e167e83643836750a68fb9691d189aaa82/src/librustc_codegen_llvm/declare.rs#L27-L30

That behaviour is inherited from LLVM. The upshot for us is that if a codegen unit defines the same symbol twice then `add_func()` will recieve the same LLVM pointer twice.

I can tell you that this doesn't happen currently, as we have an assertion that would crash out if we tried to insert into the same slot in our hash table:
https://github.com/vext01/ykrustc/blob/c42235b1b39d515ec4cb619a8923504c8d6cd63d/src/librustc/sir.rs#L75-L76

What would you like to do? We can either leave the assertion for educational purposes, or we can allow the insertion to overwrite the slot.

ltratt: I have no opinions on what we should do: I just want to make sure that a) we've thought about it b) where documentation / asserts are needed, we have added them. I'm still unsure about b) in this instance.

src/rustllvm/RustWrapper.cpp
ptersilie: I don't remember if there were others, though it wouldn't surprise me if there are.

src/librustc/sir.rs
vext01: What we currently have is conservative, in that, in doubt, it crashes. I think that's a good starting point.

ltratt: OK. So then my question is: do we need to document this, or is the `assert` enough? I have no opinion on this, but just want to make sure that we've thought about it!

vext01: I'll leave the assert, but add a brief comment.

ltratt: OK.

src/librustc_codegen_ssa/back/link.rs
vext01: Turns out the upstream comments I was thinking of were about slightly different linking aspects, so I will just add a comment.

ltratt: OK.

vext01:
I think that covers everything.

src/librustc/sir.rs
```diff
@@ -101,21 +102,20 @@ impl SirCx {
...
-        // To start, all pointer addresses are None. We would have use overwritable null pointers,
-        // but Rust won't let us and the pointee is not sized.
+        // We start with a data structure where all LLVM block pointers are unknown (None).
         let mut res = IndexVec::from_elem_n(IndexVec::default(), self.funcs.len());
         for (func_idx, func) in self.funcs.iter_enumerated() {
             res[func_idx] = IndexVec::from_elem_n(None, func.blocks.len());
         }
 
-        // Now populate the pointers.
+        // Now we populate iterate over our hash table, replacing the aforementioned `None`s.
```
ltratt: "Now we populate iterate" doesn't parse?

vext01: Fixed.

ltratt:
I think we're ready to squash? If so, please do so.

vext01:
That's the squash done.

Also the first time I have used `Co-authored-by` in the commit message. Seems to have worked.

Let's meet tomorrow to do the 3-phase merge.

ltratt:
Please feel free to push the breaker commit when you're ready.

vext01:
Cycle breaker in place.

vext01:
I think this is going to fail because buildbot is testing both the hardware and software tracer. We should probably disabled software tracing for now?

ltratt:
Fine with me.

bors r-

vext01:
bors try

ltratt:
I assumed you were going to add a commit disabling software tracing?

vext01:
> I assumed you were going to add a commit disabling software tracing?

It's a buildbot config change, so it won't appear here.

ltratt:
So why don't I r+ this, so that if it succeeds we don't build twice?

vext01:
Seems to have worked:
image

So I will stop this try job and you can r+ it.

bors try-

vext01:
> So why don't I r+ this, so that if it succeeds we don't build twice?

I just wanted to check that only one build job was submitted and I figured I could to that without bugging you ;)

vext01:
That's amusing. The cycle breaker broke the dependency checker!

vext01:
OK to squash a fix into the cycle checker commit?

ltratt:
Please do.

vext01:
Actually, the fix needs to go in the main commit, as we don't want to revert it.

I'll push separately so you can see it.

(I'm not sure why this didn't trigger before)

ltratt:
OK.

vext01:
Pushed. And I'll squash that on your word.

ltratt:
Please squash.

vext01:
Done.

vext01:
We need to disable the required swt status on github.

vext01:
I've done that, but it has confused bors.

We can probably redo the r+

vext01:
(Or if you are happy [the tests passed], you can manually merge)

vext01:
I'm not sure what's happening here... Why didn't bors merge this?

vext01:
I always forget to update bors.toml.

bors.toml
```diff
@@ -1,6 +1,6 @@
 # The service providing the commit statuses to GitHub.
 status = [
-  "buildbot/ykrustc-buildbot-build-script-hwt",
+  #"buildbot/ykrustc-buildbot-build-script-hwt",
```
ltratt: Doesn't this disable the HWT and not the SWT?

vext01: Doh. Yes. I see it's going to be one of those days!

vext01: Fixed, but I should squash.

ltratt:
Please squash.

vext01:
squashed

PR #81, opened 2020-01-15
Add YK return labels during cleanup

ptersilie:
The return labels we added initally are not sufficient as there are
multiple places in the codegen that generated function calls. One
location we missed happens within a cleanup section. This section
creates unwind blocks to catch variables that are dropped in another
process during the execution of the function call.

Adding return label generation to this section however causes compiling
of `real_drop_in_place` to segfault since it is not a real function. We
thus add an exception to skip label generation in this case.


vext01:
LGTM!

ptersilie:
bors try

ptersilie:
@vext01 Fixed. Let me know if I can squash.

vext01:
Please squash.

ptersilie:
Squashed!

PR #85, opened 2020-02-26
Start work on local variables and instructions.

vext01:
I know @ptersilie is mostly away, so assigning @ltratt for code review, and @ptersilie for a brief "concept review" if he gets a moment.

**This is a draft. There is an unfinished ykpack change to come**

EDIT: Companion PR: https://github.com/softdevteam/yk/pull/46

Commit message follows:

Starting with store and load, as these are among the simplest.

Store instructions are simple, as they don't generate a value, but load
requires a little leg-work:

The way LLVM API works, when an instruction which generates a result is
built, an implicit assignment is created behind the scenes.

For example, to emit a load from C, you'd do: `LLVMBuildLoad(x, "res")`,
where `x` is an LLVM pointer representing where to load from, and  "res"
is the name of the result. This is equivalent to an LLVM bitcode
statement `%res = load(%x)`. Put differently, each computation adds a
new local variable to the function you are emitting instructions into.

Further, the call to `LLVMBuildLoad` returns another LLVM pointer
representing the variable which will store the result of the load. To
use the result as an operand for a further computation, you'd pass the
this pointer as an argument to another API call.

E.g.
```
LLVMValueRef *a = LLVMBuildLoad(x, "a");
LLVMValueRef *b = LLVMBuildLoad(y, "y");
LLVMValueRef *c = LLVmBuildAdd(a, b, "c");
```

Note that each ValueRef in the above also identifies the instruction
which generates the value.

The upshot of all of this is that in order to build SIR instructions
from within the LLVM codegen we need to be able to identify the results
of SIR computations from LLVM ValueRef pointers.

To this end, we add a `Local` variant to `sir::Value` and when a `load`
instruction is emitted, we create a SIR load statement and a SIR local
variable for the result, but also cache the LLVM pointer.

This mechanism will be used for all instructions which generate a value,
not just for loads.

Future work: Currently every load operand is assumed to be a local
variable, but it could be something else, like a constant, or a static.

### DIFF
--- a/src/librustc/sir.rs
+++ b/src/librustc/sir.rs
@@ -48,6 +48,12 @@ newtype_index! {
     }
 }
 
+impl Into<ykpack::Local> for SirLocalIdx {
+    fn into(self) -> ykpack::Local {
+        ykpack::Local(self.as_u32())
+    }
+}
+
 /// Sir equivalents of LLVM values.
 #[derive(Debug)]
 pub enum SirValue {
@@ -235,7 +241,7 @@ impl SirCx {
         result: SirLocalIdx,
         rhs: ykpack::Rvalue,
     ) {
-        self.emit(builder, ykpack::Statement::Assign(result.as_u32(), rhs));
+        self.emit(builder, ykpack::Statement::Assign(result.into(), rhs));
     }
 
     pub fn emit_load(&mut self, builder: *const Builder, result: *const Value, arg: *const Value) {
@@ -246,7 +252,7 @@ impl SirCx {
         let res_local = self.get_or_add_local(func_idx, result).1;
         self.emit(
             builder,
-            ykpack::Statement::Assign(res_local.as_u32(), ykpack::Rvalue::Load(arg_local.as_u32())),
+            ykpack::Statement::Assign(res_local.into(), ykpack::Rvalue::Load(arg_local.as_u32())),
         );
     }
 
--- a/src/librustc/sir.rs
+++ b/src/librustc/sir.rs
@@ -5,6 +5,7 @@
 
 use rustc_data_structures::fx::FxHashMap;
 use rustc_index::{newtype_index, vec::IndexVec};
+use std::collections::hash_map::Entry;
 use std::default::Default;
 use std::io::{self, Write};
 use ykpack;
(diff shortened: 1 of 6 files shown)


src/librustc/sir.rs
```diff
@@ -164,11 +216,47 @@ impl SirCx {
...
+        &mut self,
+        builder: *const Builder,
+        result: SirLocalIdx,
+        rhs: ykpack::Rvalue,
+    ) {
+        self.emit(builder, ykpack::Statement::Assign(result.as_u32(), rhs));
+    }
+
+    pub fn emit_load(&mut self, builder: *const Builder, result: *const Value, arg: *const Value) {
+        // FIXME deal with volatile loads.
```
ltratt, line 248: Can we add an `assert` that will fail if we encounter such a load?

vext01: We won't encounter one until we have added a call from the correct place in the LLVM backend. Somewhere in/around here:
https://github.com/softdevteam/ykrustc/blob/043bf15b2c4fd0b21c02f6fe3d82bb8f663faf75/src/librustc_codegen_llvm/builder.rs#L446

ltratt: That's not my point: when we do add it somewhere else, it's *much* better if every other point we need to touch explicitly crashes. Now I don't know if we can easily tell a volatile vs. a non-volatile load here, but if we can easily do so, let's change the `FIXME` to an `assert`.

vext01: This raises a good point actually. Due to the way the LLVM codegen interacts with the SirCx, SIR's view of the world will be incomplete until we have identified and handled all of the places LLVM instructions are generated.

I think that's just going to have to be the way it is I'm afraid. There's no one 'insert an instruction' place that we can hook in to. They are scattered around the codegen.

ltratt: That still seems very orthogonal to this `FIXME` :)

vext01: (sorry github wasn't updating, so my above response looks very out of context -- I thought I was adding additional context to my original response)

> it's much better if every other point we need to touch explicitly crashes

The compiler needs to be able to build itself in order for development to continue, so I don't think we can make it crash.

vext01: We resolved this offline and decided we would identify all of the places where a Sir statement should be added and have them insert an "unknown" instruction.

This allows the compiler to build itself, but for us to know (at runtime) if Sir functionality is incomplete.

ltratt: This might be impractical/impossible, but if the "insert unknown instruction" thing (macro?) can also record where it was inserted (i.e. the source rust file), that might help us debugging in the future?

vext01: I was going to put a string inside the instruction which identifies what kind of instruction wasn't handled. That's what we did before anyway.

https://github.com/softdevteam/yk/blob/7b4cfe28cf0148813ee8efd9b58c3c023321eafd/ykpack/src/types.rs#L175-L177

(The comment is wrong to talk about MIR, will fix).

ltratt: If a macro tells you the source file/line even better, but otherwise we can add it in manualy.

vext01: There are macros in std to get those, yeah, but it's actually more convenient to have the name of the instruction in there.

Picture the scenario. You have a mir dump open in vim and see `$3 = unimplemented(file=blah.rs, line=456)`.

I'm more interested in what kind of instruction it is, and I'd have to go digging in the source to find out.

Thoughts?

ltratt: If you have `blah.rs:456` you can look it up in your editor in about 5 seconds. If it says "Add" you'll probably have 10 locations to grep for.

Of course, you can do both. `unknow_instr!(add)` can say "add" and `blah.rs:456`.

vext01: I was going to say, let's do both, but I've just noticed that quite a lot of hook points are already inside macros. I'm worried this is going to get scary.

I'll give it a go, but reserve the right to wimp out :)

ltratt: I'm sure it'll be OK, but let's see.

PR #87, opened 2020-03-10
SIRv3.

vext01:
This change moves SIR generation from `librustc_codegen_llvm` to
`librustc_codegen_ssa`.

This means that SIR will be generated from *monomorphised* MIR, rather
than from LLVM API calls. This is easier to implement and makes it
easier to access high-level program information, but means that our
notion of a block is a little further away from what ends up in the
binary. We've discussed this and think that it probably isn't going to
be an issue.

For now all SIR statements and terminators are dummy "unimplemented"
place-holders. Next we will start serialising them properly.

Co-authored-by: @ptersilie

### DIFF
--- a/src/librustc/sir.rs
+++ b/src/librustc/sir.rs
@@ -1,174 +1,114 @@
 //! Serialised Intermediate Representation (SIR).
 //!
-//! SIR is built in-memory during LLVM code-generation, and finally placed into a dedicated ELF
-//! section at link time.
+//! SIR is built in-memory during code-generation (in rustc_codegen_ssa), and finally placed
+//! into an ELF section at link time.
 
-use rustc_data_structures::fx::FxHashMap;
-use rustc_index::{newtype_index, vec::IndexVec};
+use crate::ty::{Instance, TyCtxt};
+use rustc::mir;
+use rustc_hir::def_id::LOCAL_CRATE;
+use rustc_session::config::OutputType;
+use rustc_span::sym;
+use std::cell::RefCell;
+use std::convert::TryFrom;
 use std::default::Default;
-use std::io::{self, Write};
+use std::io;
 use ykpack;
 
-// Duplicates of LLVM types defined elsewhere, copied to avoid cyclic dependencies. Whereas the
-// LLVM backend expresses pointers to these using references, we use raw pointers so to as avoid
-// introducing lifetime parameters to the SirCx (and thus into TyCtxt and every place that uses
-// it).
-extern "C" {
-    pub type Value;
-}
-extern "C" {
-    pub type BasicBlock;
-}
-extern "C" {
-    pub type Builder;
+const BUILD_SCRIPT_CRATE: &str = "build_script_build";
+
+/// A collection of in-memory SIR data structures to be serialised.
+/// Each codegen unit builds one instance of this which is then merged into a "global" instance
+/// when the unit completes.
+#[derive(Default)]
+pub struct Sir {
+    pub funcs: RefCell<Vec<ykpack::Body>>,
 }
 
-newtype_index! {
-    pub struct SirFuncIdx {
-        DEBUG_FORMAT = "SirFuncIdx({})"
+impl Sir {
+    /// Returns `true` if we should collect SIR for the current crate.
+    pub fn is_required(tcx: TyCtxt<'_>) -> bool {
+        (tcx.sess.opts.cg.tracer.encode_sir()
+            || tcx.sess.opts.output_types.contains_key(&OutputType::YkSir))
+            && tcx.crate_name(LOCAL_CRATE).as_str() != BUILD_SCRIPT_CRATE
     }
-}
 
-// The index of a block within a function.
-// Note that these indices are not globally unique. For a globally unique block identifier, a
-// (SirFuncIdx, SirBlockIdx) pair must be used.
-newtype_index! {
-    pub struct SirBlockIdx {
-        DEBUG_FORMAT = "SirBlockIdx({})"
+    /// Returns true if there is nothing inside.
+    pub fn is_empty(&self) -> bool {
+        self.funcs.borrow().len() == 0
     }
-}
 
-/// Sir equivalents of LLVM values.
-#[derive(Debug)]
-pub enum SirValue {
-    Func(SirFuncIdx),
-}
-
-impl SirValue {
-    pub fn func_idx(&self) -> SirFuncIdx {
-        let Self::Func(idx) = self;
-        *idx
+    /// Merges the SIR in `other` into `self`, consuming `other`.
+    pub fn update(&self, other: Self) {
+        self.funcs.borrow_mut().extend(other.funcs.into_inner());
     }
-}
 
-pub struct SirCx {
-    /// Maps an opaque LLVM `Value` to its SIR equivalent.
-    pub llvm_values: FxHashMap<*const Value, SirValue>,
-    /// Maps an opaque LLVM `BasicBlock` to the function and block index of its SIR equivalent.
-    pub llvm_blocks: FxHashMap<*const BasicBlock, (SirFuncIdx, SirBlockIdx)>,
-    /// Function store. Also owns the blocks
-    pub funcs: IndexVec<SirFuncIdx, ykpack::Body>,
-    /// Mirrors the insertion point for each LLVM `IrBuilder`.
-    pub builders: FxHashMap<*const Builder, (*const BasicBlock, usize)>,
-}
-
-impl SirCx {
-    pub fn new() -> Self {
-        Self {
-            llvm_values: Default::default(),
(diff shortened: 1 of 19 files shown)


src/librustc/sir.rs
```diff
@@ -1,174 +1,112 @@
...
 
-newtype_index! {
-    pub struct SirFuncIdx {
-        DEBUG_FORMAT = "SirFuncIdx({})"
+impl Sir {
+    /// Returns `true` if we should collect SIR for the current crate.
+    pub fn is_required(tcx: TyCtxt<'_>) -> bool {
+        (tcx.sess.opts.cg.tracer.encode_sir()
+            || tcx.sess.opts.output_types.contains_key(&OutputType::YkSir))
+            && tcx.crate_name(LOCAL_CRATE).as_str() != "build_script_build"
```
ltratt: Should `build_script_build` be a constant?

src/librustc/sir.rs
```diff
@@ -1,174 +1,112 @@
...
+    /// Returns true if there are no basic blocks.
+    pub fn is_empty(&self) -> bool {
+        self.func.blocks.len() == 0
     }
 
-    /// Set the current position of builder to `pos`. Equivalent to LLVMPositionBuilderBefore.
-    pub fn position_before(&mut self, builder: *const Builder, bb: *const BasicBlock, pos: usize) {
-        self.builders.insert(builder, (bb, pos));
+    /// Appends a statement to the specified basic block.
+    fn append_stmt(&mut self, bb: ykpack::BasicBlockIndex, stmt: ykpack::Statement) {
```
ltratt: `push_stmt` would be a bit more Rusty-y a name?

src/librustc/sir.rs
```diff
@@ -1,174 +1,112 @@
...
+        &mut self,
+        bb: ykpack::BasicBlockIndex,
+        mir_term: &mir::Terminator<'_>,
+    ) {
+        let term = &mut self.func.blocks[usize::try_from(bb).unwrap()].term;
+        // We should only ever replace the default unreachable terminator assigned at allocation time.
+        debug_assert!(*term == ykpack::Terminator::Unreachable);
+
+        // FIXME: Nothing is implemented yet.
+        *term = ykpack::Terminator::Unimplemented(format!("{:?}", mir_term.kind));
```
ltratt, line 106: This looks like a candidate for `file!` and `line!` to differentiate it from the case in the function below?

src/librustc_codegen_llvm/builder.rs
```diff
@@ -153,35 +146,46 @@ impl BuilderMethods<'a, 'tcx> for Builder<'a, 'll, 'tcx> {
...
     }
 
-    fn add_yk_block_label(&mut self, block: &'ll BasicBlock, lbl_name: CString) {
-        let dbg_cx = self.cx().dbg_cx.as_ref().unwrap();
-        let di_bldr = dbg_cx.get_builder();
-        unsafe {
-            llvm::LLVMRustAddYkBlockLabel(self.llbuilder, di_bldr, block, lbl_name.as_ptr());
+    fn add_yk_block_label(&mut self, fname: &str, sym: &SymbolName, bbidx: usize) {
+        if !self.tcx.sess.opts.cg.tracer.sir_labels() {
+            // We are not in hardware tracing mode.
```
ltratt, line 159: The `if` condition doesn't make it seem like we're in/not in HW tracing mode?

src/librustc_codegen_llvm/builder.rs
```diff
@@ -153,35 +146,46 @@ impl BuilderMethods<'a, 'tcx> for Builder<'a, 'll, 'tcx> {
...
+        let lbl_name = CString::new(format!("{}:{}:{}", BLOCK_LABEL_PREFIX, sym, bbidx)).unwrap();
+
+        if let Some(dbg_cx) = self.cx().dbg_cx.as_ref() {
+            let di_bldr = dbg_cx.get_builder();
+            unsafe {
+                llvm::LLVMRustAddYkBlockLabel(
+                    self.llbuilder,
+                    di_bldr,
+                    self.llbb(),
+                    lbl_name.as_ptr(),
```
ltratt, line 198: Does `LLVMRustAddYkBlockLabel` copy the contents pointed to, or do we need to leak `lbl_name` or ...?

src/librustc/sir.rs
vext01: Arguably upstream should have defined one, but (last time I checked) haven't.  We can add one for our stuff if you like, but I don't want to make changes to other parts of the compiler.

Thoughts?

src/librustc/sir.rs
vext01: Agreed.

src/librustc/sir.rs
ltratt: That string doesn't appear in the main Rust repo; anyway, it's not a big deal either way.

src/librustc/sir.rs
vext01: There's no need here as one pushes a statement, and the other sets a terminator. It's easy enough to see at a glance without the need for line numbers.

I have another branch lined up where I do use that trick for terminators. As soon as I started serialising for real I realised that you can't serialise all terminators at one call site, like I have done here.

src/librustc_codegen_llvm/builder.rs
vext01: We only need to output labels in hardware tracing mode, so it's equivalent to "are we in hardware tracing mode".

```
 150     /// Does this mode require DILabels for SIR mapping?                                                                                            
 151     pub fn sir_labels(&self) -> bool {                                                                                            
 152         match self {                                                                                            
 153             TracerMode::Hardware => true,                                                                                            
 154             TracerMode::Software | TracerMode::Off => false,                                                                                            
 155         }                                                                                            
 156     }
```

src/librustc/sir.rs
ltratt: OK.

src/librustc_codegen_llvm/builder.rs
ltratt: OK.

src/librustc_codegen_llvm/builder.rs
vext01: It's a good question.

The pointer gets passed directly into LLVM's `DIBuilder::createLabel()`. I tried to follow the call down to see what happens, but I quickly got lost in a tangle of C++.

Lots of LLVM API functions accept `char *` arguments for names of things. I'd hope that the API takes a copy, or the API would be kind of absurd (?). The user would have to ensure that the strings stay alive.

What do you think?

ltratt: I think we need to be sure about this, otherwise we're going to create code that's likely to be either wrong or inefficient.

src/librustc/sir.rs
vext01: I don't mind defining a constant tbh.

src/librustc_codegen_llvm/builder.rs
vext01: OK. After trawling the llvm sources some, I've figured it out.

[](url)

LLVM uses [StringMaps](https://llvm.org/doxygen/classllvm_1_1StringMap.html) to cache strings.

When we  do `di_builder->createLabel(..., labelName)`, the name is eventually passed to a `StringMap` responsible for caching metadata (debug info) strings. [try_emplace()](https://llvm.org/doxygen/classllvm_1_1StringMap.html#a3f1b153d947276395158efe9e5aa7693) is called and if there's a cache miss, the [string is memcpy()d into the cache](https://llvm.org/doxygen/StringMap_8h_source.html#l00197).

If the entry is already in the cache, the old value is used.

Since the string is copied into the cache, we don't have to ensure that strings in our arguments to API calls survive.

vext01:
I think this is ready for re-review.

src/librustc_codegen_llvm/builder.rs
ltratt: OK, cool!

ltratt:
Please squash.

vext01:
splat

vext01:
That last push fixes the tests. Shall I squash it in?

src/librustc_codegen_llvm/builder.rs
```diff
@@ -164,6 +164,18 @@ impl BuilderMethods<'a, 'tcx> for Builder<'a, 'll, 'tcx> {
             // Skip rustc crates since we will never trace them.
             return;
         }
+
+        // Auto-generated testing entry points causes our label insertion to crash. We don't need
```
ltratt: "causes" -> "cause"

vext01:
Typo fixed.

ltratt:
Please squash.

vext01:
Splat.

vext01:
Looks like bors crashed. Can you retry?

PR #88, opened 2020-03-11
Lower MIR call terminators to their SIR equivalent.

vext01:
The target function must be monomorphised. We had two options:

 - Duplicate monomorphisation logic in codegen_block().

 - Lower terminators deeper in the codegen and reuse existing
   monomorphisation logic.

I don't like the idea of having to keep monomorphisation logic in-sync
with upstream, so I've opted for the latter. This strategy means we now
have lots of places where we have to lower a terminator, whereas before
we had only one.

I can live with this, however I did add a macro to help us identify the
many (currently) unimplemented terminator insertions.

### DIFF
--- a/src/librustc/sir.rs
+++ b/src/librustc/sir.rs
@@ -93,17 +93,11 @@ impl SirFuncCx {
     }
 
     /// Sets the terminator of the specified block.
-    pub fn codegen_terminator(
-        &mut self,
-        bb: ykpack::BasicBlockIndex,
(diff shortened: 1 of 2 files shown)


ptersilie:
Typo in the commit message: unneeded `a` in the first line.

vext01:
If that's all, I'll do a force push to fix.

ptersilie:
Please do.

vext01:
Fixed.

PR #90, opened 2020-03-12
Abort strategy

vext01:
For simplicity, we have decided to only ever use the abort strategy for our tracer.

The second commit does that. The first moves some semi-related code around to a more suitable place and tests it.


ptersilie:
Looks good to me.

bors r+

PR #93, opened 2020-03-17
Generate Call terminators with a return value and arguments.

vext01:
Includes a fair amount of churn due to necessary ykpack changes and for
modularity.

E.g. Make lower_place() return ykpack::Place, not ykpack::Local.

Companion PR:
https://github.com/softdevteam/yk/pull/50

We will need a 3-stage merge.

### DIFF
--- a/src/librustc/sir.rs
+++ b/src/librustc/sir.rs
@@ -120,40 +120,36 @@ impl SirFuncCx {
         lvalue: &mir::Place<'_>,
         rvalue: &mir::Rvalue<'_>,
     ) -> ykpack::Statement {
-        match self.lower_place(lvalue) {
-            Ok(lhs) => {
-                let rhs = self.lower_rvalue(rvalue);
-                ykpack::Statement::Assign(ykpack::Place::from(lhs), rhs)
+        let lhs = self.lower_place(lvalue);
+        let rhs = self.lower_rvalue(rvalue);
+        ykpack::Statement::Assign(lhs, rhs)
+    }
+
+    pub fn lower_operand(&self, operand: &mir::Operand<'_>) -> ykpack::Operand {
+        match operand {
+            mir::Operand::Copy(place) | mir::Operand::Move(place) => {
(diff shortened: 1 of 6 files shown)


vext01:
(Tests still running, but we can review in parallel)

vext01:
tests passing.

src/librustc/sir.rs
```diff
@@ -120,64 +120,66 @@ impl SirFuncCx {
...
+            projection: place
+                .projection
+                .iter()
+                .map(|p| ykpack::PlaceElem::Unimplemented(format!("{:?}", p)))
+                .collect(),
         }
     }
 
-    fn lower_local(&self, local: mir::Local) -> ykpack::Local {
+    pub fn lower_local(&self, local: mir::Local) -> ykpack::Local {
```
ptersilie: Any reason why some of these lowerings are public and others aren't? I feel like it should be the same for all.

vext01: In some cases I needed access from `codegen_ssa`, but I'll double check them all.

ptersilie: Ah I see. I wonder if we should just make them all public then. It seems weird for some to be public and others not.

vext01: `lower_local()` can be private, but the others need to remain public. Change pushed.

ptersilie: So you don't think we should just make them all public? It seems weird for some to be public and others not.

ptersilie: @vext01 ping

vext01: Having things private where possible is a good thing I reckon. It stops people calling things they don't need to or shouldn't.

ptersilie: Since this is the same codebase anyways I guess it doesn't matter too much for now.

ptersilie:
Please squash and start 3-way merge.

vext01:
squashed and added cycle breaker.

ready for bors.

PR #97, opened 2020-03-23
Add binary operation lowerings.

vext01:


src/librustc/sir.rs
```diff
@@ -200,4 +213,25 @@ impl SirFuncCx {
...
+    ) -> ykpack::Rvalue {
+        let sir_op = binop_lowerings!(
+            op, Add, Sub, Mul, Div, Rem, BitXor, BitAnd, BitOr, Shl, Shr, Eq, Lt, Le, Ne, Ge, Gt,
+            Offset
+        );
+        let sir_opnd1 = self.lower_operand(opnd1);
+        let sir_opnd2 = self.lower_operand(opnd2);
+
+        if checked {
+            ykpack::Rvalue::CheckedBinaryOp(sir_op, sir_opnd1, sir_opnd2)
```
bjorn3, line 240: FYI only `Add`, `Sub`, `Mul`, `Shl` and `Shr` can be checked.

vext01: 87f32fd adds an assertion. Do you think that's good enough?

vext01:
I think this is ready to go.

ltratt:
As a general point (not really for this PR), is it a bit verbose if so many functions start with `lower_`? In my experience, a shorter prefix is more sustainable in the long term (e.g. `l_`).

vext01:
I prefer `lower_` I'm afraid.

vext01:
Let me expand on this, as I do feel somewhat strongly.

 * I can't count the number of times I've had to ask what the meaning of a single- (or two-) letter prefix is. I'm in the position where I can easily ask what they mean, but not everyone has that luxury. It's better if the code speaks for itself, leaving no ambiguity.

 * It's not the Rust way and the precedent in the compiler is to use descriptive names, e.g. `codegen_block` instead of `c_block`.

ltratt:
We've agree offline to leave the naming convention as-is in this repository.

ptersilie:
Looks fine. Please squash.

As for the `lowering_` discussion, I'm leaning towards Edd here. It's still short enough to not offend my eyes. If the function name was super longer then that would need another discussion.

vext01:
splat.

PR #99, opened 2020-04-07
Update lock file due to changes in yk.

vext01:
I was getting deserialisation errors. Took me a while to figure out what had happened.

Because we no-longer do 3-phase merges, we have no inherent lock file syncing. So we have to do it manually.

Here's one which fixes my derserialisation problem.


ptersilie:
Hm, that's also not great, though it's still slightly better than doing a 3-phase merge.

bors r+

ptersilie:
> LLVM ERROR: out of memory

Probably got unlucky.

bors r+

PR #111, opened 2020-07-22
Yk types

vext01:
Right now, only structs are supported. Everything else is serialised as an unknown type.

One design choice we should think about: in rustc, type-wise a struct is an enum with one variant. I've deviated and opted to use a separate type for structs and enums in SIR. Thoughts?

Companion PR coming soon.

### DIFF
--- a/src/librustc_codegen_llvm/sir.rs
+++ b/src/librustc_codegen_llvm/sir.rs
@@ -32,6 +32,14 @@ pub fn write_sir<'tcx>(tcx: TyCtxt<'tcx>, sir_llvm_module: &mut ModuleLlvm) {
     let mut buf = Vec::new();
     let mut encoder = ykpack::Encoder::from(&mut buf);
 
+    // First we serialise the types which will be referenced in the body packs that will follow.
+    // The serialisation order matters here, as the load order (in the runtime) corresponds with
+    // the type indices, hence use of `IndexMap` for insertion order.
+    let types =
+        tcx.sir_types.borrow_mut().map.drain(..).map(|(k, _)| k).collect::<Vec<ykpack::Ty>>();
+    let crate_hash = tcx.crate_hash(LOCAL_CRATE).as_u64();
+    encoder.serialise(ykpack::Pack::Types(ykpack::Types { crate_hash, types })).unwrap();
+
     for func in sir_funcs {
         encoder.serialise(ykpack::Pack::Body(func)).unwrap();
     }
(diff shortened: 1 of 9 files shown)


src/librustc_codegen_llvm/sir.rs
```diff
@@ -32,6 +32,14 @@ pub fn write_sir<'tcx>(tcx: TyCtxt<'tcx>, sir_llvm_module: &mut ModuleLlvm) {
     let mut buf = Vec::new();
     let mut encoder = ykpack::Encoder::from(&mut buf);
 
+    // First we serialise the types which will be referenced in the body packs that will follow.
+    // The serialisation order matters here, as the load order (in the runtime) corresponds with
+    // the type indices, hence use of `IndexMap` for insertion order.
+    let types =
+        tcx.sir_types.borrow_mut().map.drain(..).map(|(k, _)| k).collect::<Vec<ykpack::Ty>>();
```
ltratt, line 39: Is `drain` the right thing here? I was perhaps expecting `into_iter`.

vext01: Wouldn't we have to consume the `tcx `to use `into_iter`? Will try it.

ltratt: I guess the question is more: do we ever need access to `sir_types` after this point? If not, it doesn't really matter which one we do.

vext01: Ah, if you try to use `into_iter` then you end up trying to move out of a `Rc` which isn't allowed.

I figured `drain` was better than putting another inner-mutable wrapper around the innards of the `Rc`. We won't need the types again once they are serialised.

ltratt: > We won't need the types again once they are serialised.

OK, then it doesn't matter. Sorry for the noise!

ptersilie:
Looks fine. We can squash after we've done the builtbot built when we remove the priming commit.

vext01:
The priming commit is wrong. Will fix and prime.

vext01:
bors try

ltratt:
Please squash.

vext01:
Not just yet. We have to merge the other PR first.

But you can bors it if you like:
https://github.com/softdevteam/yk/pull/87

vext01:
@ptersilie OK to undo the priming commits?

ltratt:
Please go ahead.

vext01:
splat. ready for bors.

PR #119, opened 2020-08-19
Add `do_not_trace` annotations.

ptersilie:
This annotation tells the tracer to ignore the annotated function and turn it into a call instead of inlining it.

Companion PR: https://github.com/softdevteam/yk/pull/96

### DIFF
--- a/src/librustc_middle/sir.rs
+++ b/src/librustc_middle/sir.rs
@@ -86,6 +86,8 @@ impl SirFuncCx<'tcx> {
                 flags |= ykpack::bodyflags::TRACE_HEAD;
             } else if tcx.sess.check_name(attr, sym::trace_tail) {
                 flags |= ykpack::bodyflags::TRACE_TAIL;
+            } else if tcx.sess.check_name(attr, sym::do_not_trace) {
+                flags |= ykpack::bodyflags::DO_NOT_TRACE;
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/librustc_middle/sir.rs
```diff
@@ -86,6 +86,8 @@ impl SirFuncCx<'tcx> {
                 flags |= ykpack::bodyflags::TRACE_HEAD;
             } else if attr.check_name(sym::trace_tail) {
                 flags |= ykpack::bodyflags::TRACE_TAIL;
+            } else if attr.check_name(sym::do_not_trace) {
```
vext01: This mechanism has changed in upstream recently, so we will get a merge conflict at some point. Depending on if we merge the upstream sync before or after.

vext01:
LGTM. Just the question about the attribute name (mentioned in the companion PR)

src/librustc_middle/sir.rs
vext01: Since the upstream sync was just merged, this'll need syncing.

ptersilie: I didn't get a conflict here.

ptersilie: Ah, but I did get an compile error.

ptersilie:
bors try

ptersilie:
Will squash away 039b6b7 alongside the priming commit before merging.

ptersilie:
This should be ready for bors now.

PR #121, opened 2020-08-27
[2/3] Remove the software tracer and move trace_inputs to yktrace

bjorn3:
Depends on https://github.com/softdevteam/yk/pull/100


src/librustc_codegen_ssa/mir/block.rs
```diff
@@ -566,7 +566,7 @@ impl<'a, 'tcx, Bx: BuilderMethods<'a, 'tcx>> FunctionCx<'a, 'tcx, Bx> {
 
         if let Some(sfcx) = &mut self.sir_func_cx {
             let opnd = match instance {
-                Some(inst) if inst.def_id() == sfcx.trace_inputs_defid => {
+                Some(inst) if bx.tcx().symbol_name(inst).name == "__yk_trace_inputs" => {
```
vext01: Would it be better to introduce an attribute for this so that we don't have to constantly do string comparisons?

bjorn3: Done

vext01:
There are some ignored tests which can also probably go.

Search for `swt_ignore`

bjorn3:
Should I also remove all `#![no_sw_trace]` attributes from tests?

vext01:
I think so. Since software tracing will be dealt with by cranelift, right?

vext01:
I think this is ready to go then? Probably without the need for a cycle breaker?

bjorn3:
Yes, or do you want me to squash de5bfe8  and 76ba1f9?

ltratt:
@bjorn3 Our approach to squashing is that, in general, the person who raised the PR decides if/what needs squashing. So, normally, it's up to you!

bjorn3:
Ok, I have squashed those commits.

PR #125, opened 2020-09-02
Serialize SIR per codegen unit instead of per crate

bjorn3:


vext01:
I've added Lukas as well, as I want to make sure he's OK with this.

src/librustc_codegen_llvm/declare.rs
```diff
@@ -92,4 +93,19 @@ impl DeclareMethods<'tcx> for CodegenCx<'ll, 'tcx> {
             if !declaration { Some(val) } else { None }
         })
     }
+
+    fn define_sir_type(&self, ty: ykpack::Ty) -> ykpack::TypeId {
+        let mut types = self.sir.as_ref().unwrap().types.borrow_mut();
+        (types.crate_hash, types.index(ty))
```
vext01, line 99: Perhaps I was expecting type IDs to be per-cgu too? No?

bjorn3: The now confusingly named `crate_hash` in `types` is different per-cgu: https://github.com/softdevteam/ykrustc/pull/125/files#diff-ac6d225f5a8ce5d7dfbd7cdc90878877R197-R199

vext01: I also wonder if we can get these into a yorick-specfic file, to minimise merge conflicts?

bjorn3: I could add a new trait.

vext01: > The now confusingly named crate_hash in types is different per-cgu

Yes, let's rename that please.

vext01: > I could add a new trait.

Ah, I wouldn't go that far. Perhaps leave them as is then.

vext01:
What's the reason behind the second commit?

bjorn3:
SIR is something codegen related, so it shouldn't be part of `rustc_middle`. Before it was necessary as `Sir` was part of `TyCtxt`, but it isn't anymore. Combined with the third commit it also makes the diff with upstream of `rustc_middle` empty.

vext01:
OK. I remember having a hell of a time juggling the dependencies when writing that code, so these simplifications are welcomed.

Let me check one thing. What was the crate hash is now a hash derived from the cgu name and the Svh.

The Svh is the same thing as the crate hash, right?

bjorn3:
Yes, that is both correct. The `crate_hash` field should probably be renamed in a future yk PR.

vext01:
> Yes, that is both correct. The crate_hash field should probably be renamed in a future yk PR.

Yes there's going to need to be a follow up PR checking all uses of the term in our ykrustc changes and in the yk repo too.

How about `cgu_hash`?

Does the yk repo still work with this change?

@ptersilie what do you think?

bjorn3:
> Does the yk repo still work with this change?

Yes, I have tested it. No changes to yk are necessary.

src/librustc_codegen_llvm/sir.rs
```diff
@@ -49,23 +54,29 @@ pub fn write_sir<'tcx>(tcx: TyCtxt<'tcx>, sir_llvm_module: &mut ModuleLlvm) {
...
 
-    let (sir_llcx, sir_llmod) = (&*sir_llvm_module.llcx, sir_llvm_module.llmod());
+    let (sir_llcx, sir_llmod) = (&*llvm_module.llcx, llvm_module.llmod());
     let llmeta = common::bytes_in_context(sir_llcx, &buf);
     let llconst = common::struct_in_context(sir_llcx, &[llmeta], false);
 
     // Borrowed from exported_symbols::metadata_symbol_name().
     let sym_name = format!(
-        "{}_{}_{}",
+        "{}_{}_{}_{}_sym",
```
ptersilie, line 63: Are we using this symbol name for the debug labels? If so we'll need to update the hardware tracer after this (see https://github.com/softdevteam/yk/blob/master/yktrace/src/hwt/mapper.rs#L163).

ptersilie: After looking closer, I don't think we do. I guess running the tests will tell us.

bjorn3: No, it uses the function symbol name.

https://github.com/softdevteam/ykrustc/blob/b86c6ecb3121f86e29a5a873a0038dfb1558d9db/src/librustc_codegen_llvm/builder.rs#L183

ptersilie:
Looks good to me.

src/librustc_codegen_llvm/declare.rs
vext01: I'm struggling to find the code where the crate hash in the type id is changed to incorporate the crate disambiguator (the link above doesn't seem to work any more). I can see where the section name (and it's symbol) are made unique, but not in the type ID. Can you point it out please?

bjorn3: That is https://github.com/softdevteam/ykrustc/pull/125/files#diff-ac6d225f5a8ce5d7dfbd7cdc90878877R199

vext01: OK, got it:
```
  198         tcx.crate_hash(LOCAL_CRATE).hash(&mut crate_hash_hasher);                                  
  199         cgu_name.hash(&mut crate_hash_hasher);
```

So crate hash + cgu name = our new hash. LGTM.

vext01:
Please squash.

bjorn3:
Splat

PR #132, opened 2020-09-25
Use structs (instead of tuples) for trace inputs.

vext01:
This allows for named fields, which is much more ergonomic.

Companion PR ~~coming~~ https://github.com/softdevteam/yk/pull/125.

### DIFF
--- a/compiler/rustc_codegen_ssa/src/mir/block.rs
+++ b/compiler/rustc_codegen_ssa/src/mir/block.rs
@@ -594,11 +594,17 @@ impl<'a, 'tcx, Bx: BuilderMethods<'a, 'tcx>> FunctionCx<'a, 'tcx, Bx> {
                             );
                         }
 
+                        let type_err = "trace inputs should be a struct";
+
                         // Check it's the right type.
                         let local_decl = &self.mir.local_decls[ret_place.local];
                         match local_decl.ty.kind {
-                            ty::Tuple(_) => (),
-                            _ => panic!("Argument to `trace_inputs()` should be a tuple."),
+                            ty::Adt(def, _) => {
+                                if def.variants.len() != 1 {
+                                    panic!(type_err);
+                                }
+                            }
+                            _ => panic!(type_err),
                         }
 
                         sfcx.func.trace_inputs_local = Some(sfcx.lower_local(ret_place.local));
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5364,7 +5364,7 @@ dependencies = [
 [[package]]
 name = "ykpack"
 version = "0.1.0"
-source = "git+https://github.com/softdevteam/yk#56223feb6580e85f7470e2e6508afe2b2afa466e"
+source = "git+https://github.com/softdevteam/yk#02f7389286425020c20a0390d047b5930683a8a9"
 dependencies = [
  "bincode",
  "fallible-iterator",


compiler/rustc_codegen_ssa/src/mir/block.rs
```diff
@@ -594,11 +594,19 @@ impl<'a, 'tcx, Bx: BuilderMethods<'a, 'tcx>> FunctionCx<'a, 'tcx, Bx> {
...
                         }
 
+                        let type_err = || {
+                            panic!("trace inputs should be a struct");
+                        };
+
                         // Check it's the right type.
                         let local_decl = &self.mir.local_decls[ret_place.local];
                         match local_decl.ty.kind {
-                            ty::Tuple(_) => (),
```
bjorn3, line 600: Why not allow both tuples and structs?

vext01: I figured "tuple structs" would suffice. What do you think?

bjorn3: A tuple is basically an anonymous struct. Allowing them both is less surprising I think and it can be nice in some cases.

vext01: Yeah, I realised that they had the same ABI, but some part of me wants to be as restrictive as possible here.

@ptersilie thoughts?

ltratt: Why do we want to be restrictive?

vext01: @bjorn3 by the way, why isn't a tuple a `ty::Adt` in rustc? There's a separate `ty::Tuple` variant.

ptersilie: I'm with Edd that I would prefer the use of either but not both. While they are essentially the same we do end up in different paths during codegen and at least for now I find it easier if I only need to care about one type.

compiler/rustc_codegen_ssa/src/mir/block.rs
```diff
@@ -594,11 +594,19 @@ impl<'a, 'tcx, Bx: BuilderMethods<'a, 'tcx>> FunctionCx<'a, 'tcx, Bx> {
                             );
                         }
 
+                        let type_err = || {
```
ltratt: I must admit that I find this *much* uglier than having two `panic`s below.

compiler/rustc_codegen_ssa/src/mir/block.rs
ltratt: OK, fair enough.

ptersilie: But I agree that we could probably allow either in the future. Though there are not many compelling reasons to use a tuple over a struct.

bjorn3: > @bjorn3 by the way, why isn't a tuple a ty::Adt in rustc? There's a separate ty::Tuple variant.

An `Adt` is only for types defined by the user using `struct`, `enum` or `union`. Unlike `Tuple` it has a `DefId` pointing to the definition. For tuples there is no `DefId` that can be used as it is not defined in any crate.

vext01: I see. Thanks!

vext01:
Try this.

ptersilie:
Please squash.

vext01:
splat.

vext01:
priming...

bors try

vext01:
reverted priming and force pushed.

PR #140, opened 2020-11-04
New IR.

vext01:
This is a new design for SIR (and therefore TIR). The new design
resolves as many projections at SIR generation time so that they don't
have to be done at runtime. By doing so, we lessen the burden on the
trace code generator and the register allocator.

This is a companion to https://github.com/softdevteam/yk/pull/140

### DIFF
--- a/compiler/rustc_codegen_ssa/src/sir.rs
+++ b/compiler/rustc_codegen_ssa/src/sir.rs
@@ -3,154 +3,27 @@
 //! SIR is built in-memory during code-generation (in rustc_codegen_ssa), and finally placed
 //! into an ELF section at link time.
 
-use crate::mir::LocalRef;
 use crate::traits::{BuilderMethods, SirMethods};
 use indexmap::IndexMap;
 use rustc_ast::ast;
 use rustc_ast::ast::{IntTy, UintTy};
-use rustc_data_structures::fx::FxHasher;
+use rustc_data_structures::fx::{FxHashMap, FxHasher};
 use rustc_hir::{self, def_id::LOCAL_CRATE};
 use rustc_middle::mir;
 use rustc_middle::ty::print::with_no_trimmed_paths;
 use rustc_middle::ty::AdtDef;
+use rustc_middle::ty::TypeFoldable;
 use rustc_middle::ty::{self, layout::TyAndLayout, TyCtxt};
 use rustc_middle::ty::{Instance, Ty};
 use rustc_span::sym;
 use rustc_target::abi::FieldsShape;
 use rustc_target::abi::VariantIdx;
 use std::cell::RefCell;
-use std::convert::TryFrom;
+use std::convert::{TryFrom, TryInto};
 use std::default::Default;
 use std::hash::{BuildHasherDefault, Hash, Hasher};
 use std::io;
 use ykpack;
-
-pub(crate) fn lower_local_ref<'a, 'l, 'tcx, Bx: BuilderMethods<'a, 'tcx>, V>(
-    tcx: TyCtxt<'tcx>,
-    bx: &Bx,
-    decl: &'l LocalRef<'tcx, V>,
-) -> ykpack::LocalDecl {
-    let ty_layout = match decl {
-        LocalRef::Place(pref) => pref.layout,
-        LocalRef::UnsizedPlace(..) => {
-            let sir_ty = ykpack::Ty::Unimplemented(format!("LocalRef::UnsizedPlace"));
-            return ykpack::LocalDecl { ty: bx.cx().define_sir_type(sir_ty) };
-        }
-        LocalRef::Operand(opt_oref) => {
-            if let Some(oref) = opt_oref {
-                oref.layout
-            } else {
-                let sir_ty = ykpack::Ty::Unimplemented(format!("LocalRef::OperandRef is None"));
-                return ykpack::LocalDecl { ty: bx.cx().define_sir_type(sir_ty) };
-            }
-        }
-    };
-
-    ykpack::LocalDecl { ty: lower_ty_and_layout(tcx, bx, &ty_layout) }
-}
-
-fn lower_ty_and_layout<'a, 'tcx, Bx: BuilderMethods<'a, 'tcx>>(
-    tcx: TyCtxt<'tcx>,
-    bx: &Bx,
-    ty_layout: &TyAndLayout<'tcx>,
-) -> ykpack::TypeId {
-    let sir_ty = match ty_layout.ty.kind() {
-        ty::Int(si) => lower_signed_int(*si),
-        ty::Uint(ui) => lower_unsigned_int(*ui),
-        ty::Adt(adt_def, ..) => lower_adt(tcx, bx, adt_def, &ty_layout),
-        ty::Array(typ, _) => ykpack::Ty::Array(lower_ty_and_layout(tcx, bx, &bx.layout_of(typ))),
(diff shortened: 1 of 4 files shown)


compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -285,18 +173,77 @@ impl SirFuncCx<'tcx> {
             mir.basic_blocks().len()
         ];
 
+        // There will be at least as many locals in the SIR as there are in the MIR.
```
ltratt, line 176: There can be more?

compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -341,102 +289,292 @@ impl SirFuncCx<'tcx> {
...
-    pub fn lower_statement(&mut self, bb: ykpack::BasicBlockIndex, stmt: &mir::Statement<'_>) {
+    pub fn lower_statement<Bx: BuilderMethods<'a, 'tcx>>(
+        &mut self,
+        bx: &Bx,
+        bb: ykpack::BasicBlockIndex,
+        stmt: &mir::Statement<'tcx>,
+    ) {
+        self.push_stmt(
+            bb,
+            ykpack::Statement::Debug(with_no_trimmed_paths(|| format!("{:?}", stmt))),
```
ltratt: Silly question perhaps, but given that `format!` is returning a `String`, this `Debug` information might end up being quite expensive?

compiler/rustc_codegen_ssa/src/sir.rs
vext01: Yes. We no longer have a one-to-one mapping between SIR and MIR locals due to the way we decompose projections. We emit intermediate stores in certain conditions, which means we may have more locals than the corresponding MIR.

compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -341,102 +289,292 @@ impl SirFuncCx<'tcx> {
...
-        match pe {
-            mir::ProjectionElem::Field(field, ..) => ykpack::Projection::Field(field.as_u32()),
-            mir::ProjectionElem::Deref => ykpack::Projection::Deref,
-            mir::ProjectionElem::Index(local) => {
-                ykpack::Projection::Index(self.lower_local(*local))
-            }
-            _ => ykpack::Projection::Unimplemented(format!("{:?}", pe)),
-        }
+    /// Wrapper for bx.layout_of() which ensures the type is first monomorphised.
+    fn layout_of<Bx: BuilderMethods<'a, 'tcx>>(&self, bx: &Bx, t: Ty<'tcx>) -> TyAndLayout<'tcx> {
```
ltratt: This function feels like the sort of hack that might confuse us: it has the same name as another function that doesn't do the same thing. Is it so bad to remove this and just say `layout_of(self.monomorphise(...))` when necessary? Do we even need to monomorphise everything?

compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -341,102 +289,292 @@ impl SirFuncCx<'tcx> {
...
+            ykpack::IPlace::Val { off, ty, .. } => {
+                *off += add;
+                *ty = self.lower_ty_and_layout(bx, &self.layout_of(bx, mirty));
+            }
+            ykpack::IPlace::Indirect { off, ty, .. } => {
+                *off += add;
+                *ty = self.lower_ty_and_layout(bx, &self.layout_of(bx, mirty));
+            }
+            ykpack::IPlace::Const { .. } => {
+                // Oddly enough this happens. It seems to be only ever with an offset of 0 and we
```
ltratt: Comment doesn't parse.

compiler/rustc_codegen_ssa/src/sir.rs
vext01: I was actually meant to strip this debugging stuff out until I can think of a less expensive way to do it. Oops.

I'll do that now.

compiler/rustc_codegen_ssa/src/sir.rs
ltratt: OK.

compiler/rustc_codegen_ssa/src/sir.rs
vext01: We do need to monomorphise everything, and tracking down missing monomorphisations is (and was) a pain. Hence this function.

ltratt: Let's give it a different name then, something like `monomorphise_and_layout` maybe?

compiler/rustc_codegen_ssa/src/sir.rs
vext01: Actually I disagree that the type doesn't change. I'll fix that too.

compiler/rustc_codegen_ssa/src/sir.rs
vext01: Fixed in 1bd0a0fb698821f6c776591f9f821ab3bbc84811

compiler/rustc_codegen_ssa/src/sir.rs
vext01: Also fixed in 1bd0a0fb698821f6c776591f9f821ab3bbc84811

Sorry the fixes are mixed.

compiler/rustc_codegen_ssa/src/sir.rs
vext01: Fixed in 98f57e97b9102848faf8d929cd7e9fc543bc7bc2, but note that there was also a change in the yk repo.

compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -553,9 +711,139 @@ impl SirFuncCx<'tcx> {
...
+            ty::Uint(ui) => self.lower_unsigned_int_ty(*ui),
+            ty::Adt(adt_def, ..) => self.lower_adt_ty(bx, adt_def, &ty_layout),
+            ty::Array(elem_ty, len) => {
+                let align = usize::try_from(ty_layout.layout.align.abi.bytes()).unwrap();
+                let size = usize::try_from(ty_layout.layout.size.bytes()).unwrap();
+                ykpack::Ty::Array {
+                    elem_ty: self.lower_ty_and_layout(bx, &self.mono_layout_of(bx, elem_ty)),
+                    len: usize::try_from(len.eval_usize(self.tcx, ty::ParamEnv::reveal_all()))
+                        .unwrap(),
+                    byte_size: size, // FIXME use SizeAndAlign
```
vext01: I also spotted this one reviewing the FIXMEs. This should be fixed too.

vext01: Fixed in 0fe6394ae110bdd0e9bfa3748ed8a8a68b6dd320 -- there's a yk change for that too.

ltratt:
I think we're ready to prime?

vext01:
Let me do the overflow FIXME first, because it may need a change to this repo too.

vext01:
@ltratt actually having looked at the fix, and knowing that the change Lukas has queued will make the fix easier, let's fix the overflow FIXME in another PR.

If that's OK with you, there's squashing on the yk side, then we can squash this and prime.

vext01:
OK to squash here too?

ltratt:
Go ahead!

vext01:
Splat.

Priming...

bors try

ltratt:
Feel free to force push a de-priming thingy when https://github.com/softdevteam/yk/pull/140 has merged.

vext01:
Splat.

PR #143, opened 2020-11-13
Better error messages for bogus #[interp_step] functions.

vext01:
Instead of just bombing out without any useful location information, we
now point out the troublesome function definition.

Also make the errors fatal. Fixes a vector out of bounds problem when the
trace-related code tries to find the trace IO.

E.g.:
```
error: The #[interp_step] function must accept only one argument
  --> /home/vext01/research/yorick/ykrustc/src/test/compile-fail/yk/interp-step.rs:8:1
   |
LL | fn f1() {}
   | ^^^^^^^
```


vext01:
OK to squash?

ptersilie:
Yup.

vext01:
splat.

PR #147, opened 2020-11-23
sync upstream

vext01:


vext01:
```
---- sys::unix::kernel_copy::tests::bench_file_to_file_copy stdout ----
thread 'sys::unix::kernel_copy::tests::bench_file_to_file_copy' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }', library/std/src/sys/unix/kernel_copy/tests.rs:71:10
---- sys::unix::kernel_copy::tests::bench_file_to_socket_copy stdout ----
thread 'sys::unix::kernel_copy::tests::bench_file_to_socket_copy' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }', library/std/src/sys/unix/kernel_copy/tests.rs:100:10
```

I'm hoping this was a fluke, as those tests just use temp dirs.

ltratt:
OK, let's try again.

bors r+

vext01:
Thanks.

Yeah, this isn't ideal. In upstream's test:
```
      let path = crate::env::temp_dir();                                                                                                                                                                            
      let source_path = path.join("copy-spec.source");                                                                                                                                                              
      let sink_path = path.join("copy-spec.sink");   
```

where:

```
  pub fn temp_dir() -> PathBuf {                                                                                                                                                                                    
      crate::env::var_os("TMPDIR").map(PathBuf::from).unwrap_or_else(|| {                                                                                                                                           
          if cfg!(target_os = "android") {                                                                                                                                                                          
              PathBuf::from("/data/local/tmp")                                                                                                                                                                      
          } else {                                                                                                                                                                                                  
              PathBuf::from("/tmp")                                                                                                                                                                                 
          }                                                                                                                                                                                                         
      })                                                                                                                                                                                                            
  }   
```

So if both Lukas and I were to concurrently test, we might clash.

ltratt:
Can we file an upstream PR to fix this? That code will cause many people problems at some point, including us!

vext01:
Yeah, I'm looking into it now.

vext01:
What's happened now is that those files in /tmp have not been cleaned up. They are owned `vext01` so the CI worker can't overwrite them.

I'll manually delete them for now.

vext01:
Done, please retry.

vext01:
[Fix for the upstream test](https://github.com/rust-lang/rust/pull/79375)

PR #150, opened 2020-12-02
Sync upstream.

vext01:
Note that we should merge https://github.com/softdevteam/hwtracer/pull/92 first.


vext01:
Try again. This is that silly race condition again.

PR #151, opened 2020-12-07
Faster label loading.

vext01:
Companion PR ~~coming soon~~ https://github.com/softdevteam/yk/pull/170

A significant cost of starting the JIT is loading our mapping labels from DWARF.

This change extracts the labels (at compile time) into a faster to load format.

Rough benchmarking of a full yk test suite run:

before:
```
	    Mean        Std.Dev.    Min         Median      Max
real        22.323      0.305       21.924      22.303      22.858
user        127.221     0.266       126.793     127.324     127.580
sys         8.258       0.734       7.281       8.002       9.156
```

after:
```
            Mean        Std.Dev.    Min         Median      Max
real        9.082       0.544       8.398       9.157       9.896
user        98.287      0.393       97.876      98.134      98.995
sys         7.054       0.595       6.009       7.202       7.789
```

ci-yk: vext01 faster-labels

### DIFF
--- a/compiler/rustc_codegen_ssa/src/sir.rs
+++ b/compiler/rustc_codegen_ssa/src/sir.rs
@@ -936,7 +936,7 @@ pub mod labels {
     }
 
     /// Walks the DWARF tree of the specified executable and extracts Yorick location mapping
-    /// labels. The labels are returned in a vector sorted by the file offset in the executable.
+    /// labels. Returns an list of labels ordered by file offset (ascending).
     fn extract_dwarf_labels(exe_filename: &Path) -> Result<Vec<ykpack::SirLabel>, gimli::Error> {
         let file = fs::File::open(exe_filename).unwrap();
         let mmap = unsafe { memmap::Mmap::map(&file).unwrap() };
@@ -976,7 +976,7 @@ pub mod labels {
                                 // We thus simply remember the subprogram's address so we can later
                                 // assign it to the first block (ending with '_0') of this
                                 // subprogram.
-                                subaddr = Some(v as u64);
+                                subaddr = Some(u64::try_from(v).unwrap());
                             } else {
                                 panic!("Error reading dwarf information. Expected type 'Addr'.")
                             }
@@ -1002,7 +1002,8 @@ pub mod labels {
                                         let (fsym, bb) = split_label_name(s);
                                         if let gimli::AttributeValue::Addr(v) = lowpc {
                                             labels.push(ykpack::SirLabel {
-                                                off: usize::try_from(v as u64).unwrap(),
+                                                off: usize::try_from(u64::try_from(v).unwrap())
+                                                    .unwrap(),
                                                 symbol_name: fsym,
                                                 bb,
                                             });
--- a/compiler/rustc_codegen_ssa/src/sir.rs
+++ b/compiler/rustc_codegen_ssa/src/sir.rs
@@ -25,7 +25,7 @@ use std::hash::{BuildHasherDefault, Hash, Hasher};
 use std::io;
 use ykpack;
 
-const BUILD_SCRIPT_CRATE: &str = "build_script_build";
+pub const BUILD_SCRIPT_CRATE: &str = "build_script_build";
 const CHECKABLE_BINOPS: [ykpack::BinOp; 5] = [
     ykpack::BinOp::Add,
     ykpack::BinOp::Sub,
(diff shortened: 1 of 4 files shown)


compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -890,3 +890,136 @@ impl SirTypes {
...
+        let mut iter = dwarf.units();
+        let mut subaddr = None;
+        while let Some(header) = iter.next()? {
+            let unit = dwarf.unit(header)?;
+            let mut entries = unit.entries();
+            while let Some((_, entry)) = entries.next_dfs()? {
+                if entry.tag() == gimli::DW_TAG_subprogram {
+                    if let Some(_name) = entry.attr_value(gimli::DW_AT_linkage_name)? {
+                        if let Some(lowpc) = entry.attr_value(gimli::DW_AT_low_pc)? {
+                            let addr = match lowpc {
```
ltratt: This might be easier as:

```rust
if let gimli::AttributeValue::Addr(v) = v {
    addr = Some(v);
} else {
    panic!("...");
}
```

compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -890,3 +890,136 @@ impl SirTypes {
...
+        let dwarf_cow = gimli::Dwarf::load(&loader, &sup_loader)?;
+        let borrow_section: &dyn for<'a> Fn(
+            &&'a [u8],
+        )
+            -> gimli::EndianSlice<'a, gimli::RunTimeEndian> =
+            &|section| gimli::EndianSlice::new(section, endian);
+        let dwarf = dwarf_cow.borrow(&borrow_section);
+        let mut iter = dwarf.units();
+        let mut subaddr = None;
+        while let Some(header) = iter.next()? {
```
ltratt, line 966: Is this just:

```rust
for header in &iter {
   ...
}
```

in disguise?

compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -890,3 +890,136 @@ impl SirTypes {
...
+            let unit = dwarf.unit(header)?;
+            let mut entries = unit.entries();
+            while let Some((_, entry)) = entries.next_dfs()? {
```
ltratt, line 969: I'm less sure about this but is this:

```rust
for (_, entry) in &mut entries.next_dfs() {
    ...
}
```

in disguise?

compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -890,3 +890,136 @@ impl SirTypes {
...
+                            subaddr = Some(addr);
+                        }
+                    }
+                } else if entry.tag() == gimli::DW_TAG_label {
+                    if let Some(name) = entry.attr_value(gimli::DW_AT_name)? {
+                        if let Some(es) = name.string_value(&dwarf.debug_str) {
+                            let s = es.to_string()?;
+                            if s.starts_with("__YK_") {
+                                if let Some(lowpc) = entry.attr_value(gimli::DW_AT_low_pc)? {
+                                    let addr = match lowpc {
```
ltratt: This looks like its only needed in the `else` branch?

compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -890,3 +890,136 @@ impl SirTypes {
...
+                                    // Ignore labels that have no address.
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        labels.sort_by_key(|l| l.off);
```
ltratt, line 1026: Is this step necessary or is it a bit of OCD? :)

compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -890,3 +890,136 @@ impl SirTypes {
...
+pub mod labels {
+    use object::{Object, ObjectSection};
+    use std::{
+        convert::TryFrom,
+        fs,
+        path::{Path, PathBuf},
+        process::Command,
+    };
+
+    fn split_symbol(s: &str) -> (String, u32) {
```
ltratt: Is `u32` forced upon us or have we chosen it?

ltratt:
This looks like a nice speed-up! I have a few minor comments, but my only real comment is an overriding one: there's a lack of high-level documentation (e.g. in a function `///` doc comment) explaining what's going on. I think I might struggle to understand what's going on if I look at this in a year's time.

compiler/rustc_codegen_ssa/src/sir.rs
vext01: I must admit, I copy-pasted that whole function blindly. Will try it.

compiler/rustc_codegen_ssa/src/sir.rs
vext01: I have a feeling there's a reason for that. Will check.

compiler/rustc_codegen_ssa/src/sir.rs
vext01: If we maintain the list in order, we can do an early bail out optimisation at runtime.

[We currently have to search a list of address ranges. This would be better as an interval tree, but that's another story].

compiler/rustc_codegen_ssa/src/sir.rs
vext01: We've inherited that from MIR. We could change it later, but there's no need at this point in time since our BB indices are a one-to-one mapping with those of MIR (at the moment).

ltratt: Fair enough.

compiler/rustc_codegen_ssa/src/sir.rs
ltratt: Let's document that in a function doc comment: that's a *very* important invariant that the function is required to maintain.

vext01:
> my only real comment is an overriding one: there's a lack of high-level documentation

It's hairy code that was written some time ago. It's just been moved from yktrace to the compiler.

I can try to comment it up, but I think @ptersilie should review those changes, since he was the original author of this DWARF code.

compiler/rustc_codegen_ssa/src/sir.rs
vext01: Yes.

The reason for `while let Some` is that [these iterators are not standard](https://docs.rs/gimli/0.23.0/gimli/read/struct.DebugInfoUnitHeadersIter.html). They return `Result`:
image

So that means you can't use a for loop.

I doubled checked the [crate examples](https://github.com/gimli-rs/gimli/blob/d36fcf11c878c8ebfb4da4cad300773869b5b5e4/examples/simple.rs#L48), and they use `while let` too.

Looking at the features of `object` there's a [`std` feature](https://github.com/gimli-rs/gimli/blob/master/Cargo.toml#L45) that pulls in `fallible-iterator` which might let us use a for loop, but is it worth it?

ltratt: I leave that up to you :)

vext01:
Comments sorted in 8cc1d23.

I left the `while let` as is.

vext01:
OK to squash?

vext01:
Give 01d328f  a shot

ltratt:
I'm not going to die on a hill over this, but I think it's very easy for people to miss "(ascending)".

vext01:
How about:

> Returns a list of labels in ascending order of file offset.

ltratt:
I tried that, but I think it's hard to interpret, because it's not necessarily immediately obvious how to relate labels and offsets. That's why I added the "i.e.". Anyway, I'm not going to bikeshed on it further: choose what you think is best and force push a squash ready for merging.

vext01:
I see what you mean. I can't think of anything better right now.

Squashed.

PR #155, opened 2020-12-14
Pre-compute layout and offsets for `yk::Body`s.

ptersilie:
In order to initialise the SIR interpreter for blackholing, we need to
allocate memory to hold all local declarations. Instead of computing the
space needed at run-time, which is slow, do it in the compilation phase.

Companion PR: https://github.com/softdevteam/yk/pull/183

ci-yk: ptersilie sir_layout

### DIFF
--- a/compiler/rustc_codegen_ssa/src/sir.rs
+++ b/compiler/rustc_codegen_ssa/src/sir.rs
@@ -18,6 +18,7 @@ use rustc_middle::ty::{Instance, Ty};
 use rustc_span::sym;
 use rustc_target::abi::FieldsShape;
 use rustc_target::abi::VariantIdx;
+use std::alloc::Layout;
 use std::cell::RefCell;
 use std::convert::{TryFrom, TryInto};
 use std::default::Default;
@@ -107,6 +108,8 @@ pub struct SirFuncCx<'tcx> {
     pub func: ykpack::Body,
     /// Maps each MIR local to a SIR IPlace.
     var_map: FxHashMap<mir::Local, ykpack::IPlace>,
+    /// Maps TypeIds to their layout.
+    layout_map: FxHashMap<ykpack::TypeId, (usize, usize)>,
     /// The next SIR local variable index to be allocated.
     next_sir_local: ykpack::LocalIndex,
     /// The compiler's type context.
@@ -190,8 +193,17 @@ impl SirFuncCx<'tcx> {
         let mut this = Self {
             instance: instance.clone(),
             mir,
-            func: ykpack::Body { symbol_name, blocks, flags, local_decls, num_args: mir.arg_count },
+            func: ykpack::Body {
+                symbol_name,
+                blocks,
+                flags,
+                local_decls,
+                num_args: mir.arg_count,
+                layout: (0, 0),
+                offsets: Vec::new(),
+            },
             var_map,
+            layout_map: FxHashMap::default(),
             next_sir_local: 0,
             tcx,
         };
@@ -213,6 +225,20 @@ impl SirFuncCx<'tcx> {
             this.next_sir_local += 1;
         }
         this
+    }
+
+    /// Compute layout and offsets required for blackholing.
+    pub fn compute_layout_and_offsets(&mut self) {
+        let mut layout = Layout::from_size_align(0, 1).unwrap();
+        for ld in &self.func.local_decls {
+            let (size, align) = self.layout_map.get(&ld.ty).unwrap();
+            let l = Layout::from_size_align(*size, *align).unwrap();
+            let (nl, off) = layout.extend(l).unwrap();
+            self.func.offsets.push(off);
+            layout = nl;
+        }
+        layout = layout.pad_to_align();
+        self.func.layout = (layout.size(), layout.align());
     }
 
     /// Returns the IPlace corresponding with MIR local `ml`. A new IPlace is constructed if we've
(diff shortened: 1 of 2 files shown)


compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -215,6 +227,20 @@ impl SirFuncCx<'tcx> {
         this
     }
 
+    /// Compute layout and offsets required for blackholing.
+    pub fn compute_layout_and_offsets(&mut self) {
+        let mut layout = Layout::from_size_align(0, 1).unwrap();
```
vext01, line 232: Why `1` here? I suppose alignment doesn't matter for zero bytes?

compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -215,6 +227,20 @@ impl SirFuncCx<'tcx> {
...
+        for ld in &self.func.local_decls {
+            let (size, align) = self.layout_map.get(&ld.ty).unwrap();
+            let l = Layout::from_size_align(*size, *align).unwrap();
+            let (nl, off) = layout.extend(l).unwrap();
```
vext01, line 236: couldn't you assign directly to `layout` in the first element of the return tuple?

compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -215,6 +227,20 @@ impl SirFuncCx<'tcx> {
         this
     }
 
+    /// Compute layout and offsets required for blackholing.
+    pub fn compute_layout_and_offsets(&mut self) {
+        let mut layout = Layout::from_size_align(0, 1).unwrap();
+        for ld in &self.func.local_decls {
+            let (size, align) = self.layout_map.get(&ld.ty).unwrap();
+            let l = Layout::from_size_align(*size, *align).unwrap();
```
vext01, line 235: I wonder if `decl_layout` would be more descriptive, since we have many layouts in scope here.

compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -107,6 +108,8 @@ pub struct SirFuncCx<'tcx> {
     pub func: ykpack::Body,
     /// Maps each MIR local to a SIR IPlace.
     var_map: FxHashMap<mir::Local, ykpack::IPlace>,
+    /// Maps TypeIds to their layout.
+    layout_map: FxHashMap<ykpack::TypeId, (usize, usize)>,
```
vext01, line 112: How about `FxHashMap<ykpack::TypeId, ykpack::SizeAndAlign>`?

compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -779,7 +802,9 @@ impl SirFuncCx<'tcx> {
             ty::Tuple(..) => self.lower_tuple_ty(bx, ty_layout),
             _ => ykpack::Ty::Unimplemented(format!("{:?}", ty_layout)),
         };
-        bx.cx().define_sir_type(sir_ty)
+        let tyid = bx.cx().define_sir_type(sir_ty);
+        self.layout_map.insert(tyid, (size, align));
```
vext01, line 806: Does it make more sense to do this inside `define_sir_type` and store the layout map inside the `Sir` object of the codegen context?

vext01: That way we don't duplicate work when using the same types in different `SirFuncCx`s

compiler/rustc_codegen_ssa/src/sir.rs
ptersilie: That's just the minimum allowed alignment: https://doc.rust-lang.org/std/alloc/struct.Layout.html#impl

compiler/rustc_codegen_ssa/src/sir.rs
ptersilie: I assume because of the `let` it would create a new variable instead of overwriting the existing one.

compiler/rustc_codegen_ssa/src/sir.rs
ptersilie: Since we immediately use this variable in the next line and then never again, I'm okay with it being short. But I could merge those two lines into one if you like.

compiler/rustc_codegen_ssa/src/sir.rs
ptersilie: Could do, but then would have to cast the values twice, once when putting it into the `layout_map`, and then again when taking them out, since `Layout` works on `usize`s.

compiler/rustc_codegen_ssa/src/sir.rs
ptersilie: I started doing it there first, but then realised putting it into `Sir` means the `layout_map` will be serialised, even though we have no need for it later on.

compiler/rustc_codegen_ssa/src/sir.rs
vext01: I see. Does `Layout::new::<()>()` work?

compiler/rustc_codegen_ssa/src/sir.rs
vext01: Ah yes. My bad.

compiler/rustc_codegen_ssa/src/sir.rs
vext01: Only if you see fit.

compiler/rustc_codegen_ssa/src/sir.rs
ptersilie: No idea. I just followed the documentation: https://doc.rust-lang.org/std/alloc/struct.Layout.html#examples

compiler/rustc_codegen_ssa/src/sir.rs
vext01: I'm not sure that's right. Looks like we have fine-control over what gets serialised:
 * https://github.com/softdevteam/ykrustc/blob/6d1940069431afc538c031f94aafd1046bb9c36a/compiler/rustc_codegen_llvm/src/base.rs#L168-L169
 * https://github.com/softdevteam/ykrustc/blob/6d1940069431afc538c031f94aafd1046bb9c36a/compiler/rustc_codegen_llvm/src/sir.rs#L42-L54

ptersilie: Ah, fair enough. I could put it into `Sir` then if you prefer, though I would have to rewrite most of this PR. Any strong argument why it's better there?

vext01: That's unfortunate :\ sorry.

The reason it's better there is because each `SirFuncCx` is for one SIR `Body` and the same types can appear in multiple SIR bodies. Therefore if we compute the per-type layouts on a per-body basis, then we end up doing the same work many times.

`define_sir_type` (potentially a bad function name actually) only defines the type if we haven't seen that type before, otherwise it just returns the cached value. So you should be able to piggy back off that logic and only generate the layout if we haven't seen that type before?

Please check my reasoning though :)

ptersilie: Sounds reasonable. Okay, I'll rewrite.

vext01: We could do it in two stages if you like? We can merge this, and then the next PR optimises it?

vext01: In fact, I think I'd prefer that in case what I proposed isn't possible for whatever reason.

ptersilie: Are you okay with me changing `define_sir_type` then, as I'd have to pass the layout as arguments. Or do you want me to add another function to the `Sir`?

ptersilie:
Okay, let's move one of the issues above into a separate PR. In that case I think there's nothing to change here, so this is ready for merging.

compiler/rustc_codegen_ssa/src/sir.rs
vext01: Yes, that's fine if you need to.

But we might be able to compute the layouts [here](https://github.com/softdevteam/ykrustc/blob/6d1940069431afc538c031f94aafd1046bb9c36a/compiler/rustc_codegen_ssa/src/sir.rs#L884). We can add your mapping to `SirTypes` without fear of it being serialized.

Only `SirTypes::map` is serialized.

compiler/rustc_codegen_ssa/src/sir.rs
vext01: Fair enough. Don't worry then.

compiler/rustc_codegen_ssa/src/sir.rs
ptersilie: I don't think we can do it here, since we don't have access to the size and align values. Though `ykpack::Ty` has `size` and `align` methods, we haven't implemented this for all Rust types, so for now we have to rely on the size and align we get from `ty_layout.layout.align/size`.

vext01:
OK. This PR will need the branch override set and we need to squash the yk side.

ptersilie:
Branch override set.

vext01:
So this is ready to go now I think?

ptersilie:
It is.

PR #157, opened 2020-12-15
Remove map storing type layouts.

ptersilie:
Since we now store size and alignments on the SIR types, we no longer
need the layout map to compute the memory layout for blackholing.

### DIFF
--- a/compiler/rustc_codegen_ssa/src/sir.rs
+++ b/compiler/rustc_codegen_ssa/src/sir.rs
@@ -108,8 +108,6 @@ pub struct SirFuncCx<'tcx> {
     pub func: ykpack::Body,
     /// Maps each MIR local to a SIR IPlace.
     var_map: FxHashMap<mir::Local, ykpack::IPlace>,
-    /// Maps TypeIds to their layout.
-    layout_map: FxHashMap<ykpack::TypeId, (usize, usize)>,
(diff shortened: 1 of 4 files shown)


compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -903,6 +899,10 @@ impl SirTypes {
             idx
         })
     }
+
+    pub fn get_value(&self, tyid: ykpack::TypeId) -> &ykpack::Ty {
```
vext01: I'd just call that `get()`.

vext01:
This looks good. I suppose there is no yk-level change?

compiler/rustc_codegen_ssa/src/sir.rs
ptersilie: I called it `get_value` in case we might want to have a function that returns the key someday. But I'm happy to change it.

ptersilie:
No changes in yk needed.

ptersilie:
Addressed the comment and also added a small doc comment.

ptersilie:
Should I squash?

vext01:
Sorry, yes, please squash.

ptersilie:
Squashed.

vext01:
and, finally....

bors r+

PR #159, opened 2020-12-18
Implement missing bits for array slice fat pointers.

vext01:
Companion to https://github.com/softdevteam/yk/pull/204

ci-yk: vext01 len-test

### DIFF
--- a/compiler/rustc_codegen_ssa/src/sir.rs
+++ b/compiler/rustc_codegen_ssa/src/sir.rs
@@ -411,6 +411,27 @@ impl SirFuncCx<'tcx> {
                 self.lower_binop(bx, bb, dest_ty, *op, opnd1, opnd2, true)
             }
             mir::Rvalue::Cast(mir::CastKind::Misc, op, ty) => self.lower_cast_misc(bx, bb, op, ty),
+            mir::Rvalue::Len(p) => {
+                let ip = self.lower_place(bx, bb, p);
+                match p.ty(&self.mir.local_decls, self.tcx).ty.kind() {
+                    ty::Array(_elem_ty, len) => {
+                        let raw_val =
+                            usize::try_from(len.eval_usize(self.tcx, ty::ParamEnv::reveal_all()))
+                                .unwrap();
+                        let val = ykpack::Constant::Int(ykpack::ConstantInt::UnsignedInt(
+                            ykpack::UnsignedInt::Usize(raw_val),
+                        ));
+                        ykpack::IPlace::Const { val, ty: ip.ty() }
+                    }
+                    ty::Slice(_elem_ty) => self.offset_iplace(
+                        bx,
+                        ip,
+                        i32::try_from(self.tcx.data_layout.pointer_size.bits()).unwrap(),
+                        dest_ty,
+                    ),
+                    _ => unreachable!(),
+                }
+            }
             _ => ykpack::IPlace::Unimplemented(with_no_trimmed_paths(|| {
                 format!("unimplemented rvalue: {:?}", rvalue)
             })),
@@ -536,8 +557,8 @@ impl SirFuncCx<'tcx> {
                         }
                     }
                 }
-                mir::ProjectionElem::Index(idx) => {
-                    if let ty::Array(elem_ty, ..) = cur_mirty.kind() {
+                mir::ProjectionElem::Index(idx) => match cur_mirty.kind() {
+                    ty::Array(elem_ty, ..) | ty::Slice(elem_ty) => {
                         let arr_lay = self.mono_layout_of(bx, cur_mirty);
                         let elem_size = match &arr_lay.fields {
                             FieldsShape::Array { stride, .. } => {
@@ -559,30 +580,38 @@ impl SirFuncCx<'tcx> {
                         self.push_stmt(bb, stmt);
                         cur_iplace = dest.to_indirect(dest_ty);
                         elem_ty
-                    } else {
+                    }
+                    _ => {
                         return ykpack::IPlace::Unimplemented(format!("index on {:?}", cur_mirty));
                     }
-                }
+                },
                 mir::ProjectionElem::Deref => {
-                    if let ty::Ref(_, ty, _) = cur_mirty.kind() {
-                        if let ykpack::IPlace::Indirect { ty: dty, .. } = cur_iplace {
-                            // We are dereffing an already indirect place, so we emit an
-                            // intermediate store to strip away one level of indirection.
-                            let dest = self.new_sir_local(dty);
-                            let deref = ykpack::Statement::Store(dest.clone(), cur_iplace.clone());
-                            self.push_stmt(bb, deref);
-                            cur_iplace = dest;
-                        }
+                    match cur_mirty.kind() {
+                        ty::Ref(_, ty, _) | ty::RawPtr(ty::TypeAndMut { ty, .. }) => {
+                            if let ykpack::IPlace::Indirect { ty: dty, .. } = cur_iplace {
+                                // We are dereffing an already indirect place, so we emit an
+                                // intermediate store to strip away one level of indirection.
+                                let dest = self.new_sir_local(dty);
+                                let deref =
+                                    ykpack::Statement::Store(dest.clone(), cur_iplace.clone());
+                                self.push_stmt(bb, deref);
+                                cur_iplace = dest;
+                            }
 
-                        if let Some(l) = cur_iplace.local() {
-                            self.notify_referenced(l);
-                        }
+                            if let Some(l) = cur_iplace.local() {
+                                self.notify_referenced(l);
+                            }
 
-                        let tyid = self.lower_ty_and_layout(bx, &self.mono_layout_of(bx, ty));
-                        cur_iplace = cur_iplace.to_indirect(tyid);
-                        ty
-                    } else {
-                        return ykpack::IPlace::Unimplemented(format!("deref non-ref"));
+                            let tyid = self.lower_ty_and_layout(bx, &self.mono_layout_of(bx, ty));
+                            cur_iplace = cur_iplace.to_indirect(tyid);
+                            ty
+                        }
+                        _ => {
+                            return ykpack::IPlace::Unimplemented(format!(
+                                "invalid deref: {:?}",
+                                cur_mirty.kind()
+                            ));
+                        }
                     }
                 }
                 _ => return ykpack::IPlace::Unimplemented(format!("projection: {:?}", pj)),


compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -411,6 +411,18 @@ impl SirFuncCx<'tcx> {
...
             mir::Rvalue::Cast(mir::CastKind::Misc, op, ty) => self.lower_cast_misc(bx, bb, op, ty),
+            mir::Rvalue::Len(p) => {
+                // Yorick's register allocator will always put a fat pointer on the stack, so
+                // finding its length is a matter of skipping to the second field of the fat
+                // pointer.
+                let ip = self.lower_place(bx, bb, p);
+                self.offset_iplace(
+                    bx,
+                    ip,
+                    i32::try_from(std::mem::size_of::<usize>()).unwrap(),
```
bjorn3: This is not correct for cross-compiling. You will need to take it from `tcx.data_layout.pointer_size.bits()`.

vext01: Thanks. Fixed in a4edfc4

compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -411,6 +411,18 @@ impl SirFuncCx<'tcx> {
                 self.lower_binop(bx, bb, dest_ty, *op, opnd1, opnd2, true)
             }
             mir::Rvalue::Cast(mir::CastKind::Misc, op, ty) => self.lower_cast_misc(bx, bb, op, ty),
+            mir::Rvalue::Len(p) => {
+                // Yorick's register allocator will always put a fat pointer on the stack, so
```
ltratt: What does "always put a fat pointer on the stack" mean? A struct that's something like `struct S {x: &dyn T}` could be on the heap.

bjorn3: `Rvalue::Len` on arrays is also valid: https://github.com/bjorn3/rustc_codegen_cranelift/blob/54ba238a6d4d1c2106848ffeb44b6932e95d27fa/src/base.rs#L916-L933 Looking at https://github.com/bjorn3/rustc_codegen_cranelift/commit/f4e544894abc6424d70ec27c196ef4d722dd9379 it seems to be done in the drop glue for `[T; N]`.

vext01: Currently the JIT doesn't understand the heap (will crash trying to compile a `Box`). It only knows that since the value is 128-bit, it can't fit in a register. You are right though, we should future proof this comment. How about:

> For now, Yorick's register allocator will always put fat pointers in memory, ...

One optimisation that could come later is to relax this constraint, and allow 128-bit values to span two-registers. But now isn't the time, I don't think.

vext01: If I understand properly, in the case of an array, we can statically insert a constant length? I think I'd like to do that as a separate PR, but here we should at least insert an "unimplemented" statement for arrays.

Sounds like a sensible plan?

ltratt: "For now, Yorick's register allocator will always put fat pointers in memory" is very easily misunderstood. Maybe this should be more of a "FIXME: currently this relies on the fact that yk's register allocator always stores fat pointers on the stack". An `assert`, if possible, would be an even better alternative!

vext01: The thing is that it's not clear if it's something we'd want to "fix" without benchmarking.

What would you be asserting? Register allocation decisions happen much later at trace compile time.

vext01: So how about:

> this relies on the fact that yk's register allocator never stores fat pointers in registers.

ltratt: > The thing is that it's not clear if it's something we'd want to "fix" without benchmarking.

I think you can safely assume that, if used in a tight loop, anything (including, but not limited to, fat pointers) will benefit from being in registers.

> What would you be asserting?

Can't we assert that it actually *has* been stored on the stack and not elsewhere?

vext01: Afraid not, because no register allocation decisions are made at SIR lowering time. Registers are allocated during trace code-generation.

ltratt: Right I see. Obvious other question, then: can we calculate its length another / better way?

vext01: As I understand, the reason that Rust does it this way is that the length of a slice can't be known statically. Otherwise there'd be no need for a fat pointer and a constant could be inserted (as it appears they do for arrays, whose length, unlike slices, is encoded in its type).

ltratt: The size of a slice is constant (two machine words). The second element tells you how many items it points to, which is rather different. I'm not sure which we're talking about now.

vext01: You lost me.

I thought you were asking if we could compute the second element of a fat pointer statically. I don't think you can.

vext01: Did you mean, can we make it so that ykrustc knows what could live in a register at runtime?

That could be done if we share `can_live_in_register()` somehow.

ltratt: > I thought you were asking if we could compute the second element of a fat pointer statically. I don't think you can.

There are two sized things here: how many bytes does the *slice* occupy? how many items does the slice point to? The latter is dynamic and thus unknowable (in general) at compile-time. The former is static and knowable at compile-time (if it wasn't, we wouldn't be able to compile anything). I'm trying to work out which of these two things we're getting the size of here.

vext01: `Rvalue::len`, the MIR bytecode, accesses the length of the collection pointed to (be it array or slice, although this change addresses the slice case).

We are lowering such a bytecode to a SIR bytecode that accesses the fat pointer + 8.

I'm starting to wonder if my original comment is a bit misguided. Even if the code-gen could span a 128-bit value over 2 registers, we could still make the `+8` work by considering two registers as a "paired" (i.e. we consider two registers (e.g.) `rax` and `rbx` as contiguous storage, and reading past `rax` leads into `rbx`).

What do you think? Does that make sense? Assuming it does, I suggest I remove the comment altogether, as in hindsight it seems to do more harm than good.

ltratt: I don't think we would necessarily put the two values in consecutive registers (and does it even make sense to view registers as storage?!).

It feels like the right comment might be "FIXME: this won't work correctly if the second element of the slice is not stored immediately after the first"?

vext01: > I don't think we would necessarily put the two values in consecutive registers.

I'm not saying the registers would need to be consecutive, just that the allocator could in theory issue a pair of registers (any two) and consider them a 128-bit space. Rust does exactly this over the ABI boundary for 128-bit values. A fat pointer is split between two registers if possible.

> does it even make sense to view registers as storage?

I'm not sure what the controversy is there :P When I `mov` a value into a register, it can be thought of as a storing bytes into the register, can it not?

ltratt: Let's maybe ignore the register issue for now :) Is the "FIXME" I suggested OK, or do we need to reconsider it?

vext01: Having reviewed this thread, I think neither my original comment nor the proposed FIXME above is right.

SIR is abstract in the sense that it makes no assumptions about register allocation decisions. The code in question asks to access 8 bytes at an 8 byte offset of an abstract 16-byte value. It just so happens that for now that (for simplicity) the register allocator will put 16-byte values on the stack. But there's no reason a 16-byte value couldn't span two (arbitrary) registers at runtime. Rust already does things like this, probably for performance reasons.

Therefore a better way forward is to remove the comment altogether, as it's confusing.

Would you be happy with that?

ltratt: To be clear: if the register allocator completely changes how it lays out the 16 byte value, nothing in this piece of code will need to change?

vext01: That's what I'm aiming at, yeah.

(It should also be possible to offset constants in the same way. Something else on my to-do list)

ltratt: Aiming at? Can we be a bit more precise please for the sake of my constitution :)

vext01: Yes, that is how I would like to implement this. 😄

ltratt: Just to be clear (because I'm still not clear): nothing in this code will need to change, no matter what we change in the register allocator? Is that the case now, or the case that we would like it to be at some point in the future?

vext01: Nothing in this code will need to change.

ltratt: OK, got it.

ltratt: So I think we're just waiting on this one?

vext01: I've implemented most of it, but we are hitting an `unimplemented!` now. Bear with me.

ltratt: OK :)

vext01: Fixed in d816a7d.

vext01: Fixed in 2fb17a4.

vext01:
Once we are happy with these changes a rebase is needed (on both repos). I've tested the rebase, all tests pass.

compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -413,12 +413,24 @@ impl SirFuncCx<'tcx> {
...
-                self.offset_iplace(
-                    bx,
-                    ip,
-                    i32::try_from(self.tcx.data_layout.pointer_size.bits()).unwrap(),
-                    dest_ty,
-                )
+                match p.ty(&self.mir.local_decls, self.tcx).ty.kind() {
+                    ty::Array(_elem_ty, len) => {
+                        let raw_val =
+                            usize::try_from(len.eval_usize(self.tcx, ty::ParamEnv::reveal_all()))
```
ltratt, line 419: Does `eval_usize` return a `usize`? [The name is suggestive!]

vext01: They suckered you in too!

https://github.com/rust-lang/rust/blob/b5c496de374407c0937cf5257d89230b5358d053/compiler/rustc_middle/src/ty/consts.rs#L197-L202

ltratt: I was going to write something sarky, but life is too short!

ltratt:
Please squash.

compiler/rustc_codegen_ssa/src/sir.rs
vext01: If it makes you feel better, this confused me too!

vext01:
Squashed and rebased.

PR #160, opened 2020-12-18
Disable debug assertions for libstd

bjorn3:
The overhead of them is huge while the likelihood of triggering them is not very big. The `is_aligned_and_not_null` checks in several unsafe libcore functions alone are responsible for 18% of the time it takes to test Yorick.

### DIFF
--- a/.buildbot.config.toml
+++ b/.buildbot.config.toml
@@ -1,8 +1,9 @@
 # Config file for continuous integration.
 
 [rust]
-codegen-units = 0           # Use many compilation units.
-debug-assertions = true     # Turn on assertions in rustc.
+codegen-units = 0            # Use many compilation units.
+debug-assertions = true      # Turn on assertions in rustc.
+debug-assertions-std = false # Assertions in libstd cause too much overhead.
 
 [build]
 docs = false
@@ -10,7 +11,7 @@ extended = true
 tools = ["cargo", "rustdoc"]
 
 [llvm]
-assertions = true           # Turn on assertions in LLVM.
+assertions = true            # Turn on assertions in LLVM.
 
 [install]
 prefix = "build/ykrustc-stage2-latest"


ltratt:
Can I ask what you profiled to discover this? The test suite I guess?

bjorn3:
Yes, I ran `RUSTFLAGS="-Ctracer=hw" perf record --call-graph dwarf cargo test` with all tests already built.

vext01:
Do we only want to do this is CI or more generally?

bjorn3:
`debug-assertions-std` defaults to the value of `debug-assertions` which itself defaults to false. I used the same `config.toml` as CI locally. I don't know if any of you do this. If you don't enable `debug-assertions`, `debug-assertions-std` isn't enabled by default either.

.buildbot.config.toml
```diff
@@ -1,16 +1,17 @@
...
+debug-assertions-std = false # Assertions in libstd cause too much overhead.
 
 [build]
 docs = false
 extended = true
 tools = ["cargo", "rustdoc"]
 
 [llvm]
-assertions = true           # Turn on assertions in LLVM.
+assertions = true            # Turn on assertions in LLVM.
```
ltratt, line 14: I think a space has been inserted accidentally?

bjorn3: It is required to align all comments to make @vext01 happy.

ltratt:
I'm not 100% sure I understand Edd's comment but I *think* he's asking the following. It seems reasonable to turn off assertions in the general case, but when we're running rustc's test suite, should we force them on?

.buildbot.config.toml
ltratt: OK, well, I don't want him to be unhappy :p

bjorn3:
Rust does run with all debug assertions enabled on all but a few builders. For ykrustc though as changes from upstream don't use unsafe code nor directly affect codegen and as such are unlikely to cause miscompilations. On the other hand I guess it could be useful for detecting some kinds of bugs in the trace compilation.

ltratt:
AIUI, this change would only affect how long CI takes to run?

bjorn3:
Yes and if you like me use the same config as CI for local development, it will make local testing faster too.

ltratt:
OK, I think I understand. Although I can be convinced otherwise, it does feel to me like we should keep the CI as it is with "maximal" checks on. [I don't use the CI config, so I haven't noticed this particular problem.]

bjorn3:
Ok, I will just disable it locally then.

PR #173, opened 2021-02-23
Add mmap2 to deps.rs.

vext01:


ltratt:
Can we get rid of mmap at the same time?

vext01:
A couple of places in rustc still use the old `mmap`. This should be fixed upstream though.

(We currently don't run security audits on ykrustc, as upstream code always upsets `cargo audit`)

ltratt:
OK.

bors r+

PR #174, opened 2021-02-23
Reject closures marked #[interp_step].

vext01:
I was investigating ways of parametrising tests. One way involved passing variants of the interpreter step as closures. I found out the hard way that this won't work.

We can't accept closures, as these have a hidden argument prepended for the upvars and this breaks the invariant that Local(1) is the interpreter context.

This change causes ykrustc to reject a program that marks a closure `#[interp_step]`.

### DIFF
--- a/compiler/rustc_codegen_ssa/src/sir.rs
+++ b/compiler/rustc_codegen_ssa/src/sir.rs
@@ -101,6 +101,18 @@ impl SirFuncCx<'tcx> {
                     );
                 }
 
+                let inst_kind = instance.ty(tcx, ty::ParamEnv::reveal_all()).kind();
+                if !matches!(inst_kind, ty::FnDef(..)) {
+                    // We reject anything that isn't a regular function to ensure that we don't
+                    // encounter odd problems elsewhere. For example, closures prepend an argument
+                    // for upvars which breaks the invariant that Local(1) is the interpreter
+                    // context.
+                    tcx.sess.span_fatal(
+                        tcx.def_span(instance.def_id()),
+                        "#[interp_step] can only be applied to regular function definitions",
+                    );
+                }
+
                 if mir.args_iter().count() != 1 {
                     tcx.sess.span_fatal(
                         tcx.def_span(instance.def_id()),
--- a/src/test/ui/yk/interp-step-closure.rs
+++ b/src/test/ui/yk/interp-step-closure.rs
@@ -1,0 +1,13 @@
+// build-fail
+// dont-check-compiler-stderr
+// compile-flags: -C tracer=hw -C opt-level=0
+
+#![feature(stmt_expr_attributes)]
+
+struct Ctx;
+
+fn main() {
+    let step = #[interp_step] |ctx: &mut Ctx| {};
+    //~^ #[interp_step] can only be applied to regular function definitions
+    step(&mut Ctx);
+}


ltratt:
Instead of rejecting closures, should we only let through things that we know for sure work? Or is the only meaningful difference "closures or not closures"?

vext01:
I did try annotating a struct, and the compiler accepted it.  It seems only function-like things make it into this particular section of code.

If we want to reject the others I'd have to dig deeper -- not something I'd like to do right now.

ltratt:
What I'm suggesting is not that we list *everything* to reject but rather only list what we want to include and reject everything else.

vext01:
That's what we are doing.

Note the `!` before `matches!`.

compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -101,6 +101,17 @@ impl SirFuncCx<'tcx> {
                     );
                 }
 
+                let inst_kind = instance.ty(tcx, ty::ParamEnv::reveal_all()).kind();
+                if !matches!(inst_kind, ty::FnDef(..)) {
+                    // We can't accept closures, as these have a hidden argument prepended for the
```
ltratt: I *think* this comment should be something like "We can't accept anything that's not a normal function definition (e.g. we can't accept closures because of their hidden upvars argument)".

vext01: That's fine by me.

vext01:
Try the second hunk of 97392dd. I bodged the commit, the first hunk needs to be removed at squash time.

bjorn3:
> If we want to reject the others I'd have to dig deeper -- not something I'd like to do right now.

All other checks of this kind are ad-hoc at https://github.com/rust-lang/rust/blob/446d4533e89db04f9568be4199e56b5fce0d176d/compiler/rustc_passes/src/check_attr.rs

compiler/rustc_codegen_ssa/src/sir.rs
```diff
@@ -103,9 +103,9 @@ impl SirFuncCx<'tcx> {
 
                 let inst_kind = instance.ty(tcx, ty::ParamEnv::reveal_all()).kind();
                 if !matches!(inst_kind, ty::FnDef(..)) {
-                    // We can't accept closures, as these have a hidden argument prepended for the
-                    // upvars and this breaks the invariant that Local(1) is the interpreter
-                    // context.
+                    // We can't accept anything that isn't a regular function. This includes
```
ltratt: I think we need to deemphasise the second sentence a bit. Something like "We reject anything that isn't a regular function to ensure that we don't encounter odd problems elsewhere. For example, closures prepend an argument for upvars which breaks the invariant that Local(1) is the interpreter context."

vext01:
I see. So we'd have to add a check in there for our stuff.

One for another day I think.

vext01:
282632d is copy-paste-format your wording :)

ltratt:
That new commit displays exceptional levels of wisdom. I couldn't have written it better myself 📦:p

Please squash.

vext01:
splat.

vext01:
There's a typo in the test file.
> #[interp_step] can only be **appled** to regular function definitions.

Ok to fix and force push?

ltratt:
Please go ahead.

vext01:
splat

PR #183, opened 2021-04-19
tracer selection flag

vext01:
The meat of this change is in the first commit. See the commit message for more.


vext01:
Let's put this on hold. We spoke offline and it may be easier to make the tracer decision happen at the llvm level, thus avoiding changes to rustc.

ltratt:
I think we should probably close this now?

